Issue #13, opened 2022-05-15
Fails to build... but still kinda works


DevinR528:
When I run the command in the readme it builds and I get `dbg!` output but I also get
```
error: failed to run custom build command for `linter_api v0.1.0 (/home/devinr/aprog/rust/__forks__/rust-linting/linter_api)`

Caused by:
  could not execute process `/home/devinr/aprog/rust/__forks__/rust-linting/target/debug/build/linter_api-aa48e93580444ad5/build-script-build` (never executed)

Caused by:
  No such file or directory (os error 2)
warning: build failed, waiting for other jobs to finish...
```
I only noticed when I tried to only `dbg!` print the `ItemType::Function`.

xFrednet:
Hmm, weird, on what crate do you try to run the linter?

xFrednet:
For the `rust-linting` repo cargo might try to execute the `build.rs` file in `linter_api`. And it looks like this fails for some reason ...

DevinR528:
Yeah I just did `cargo r --bin cargo-linter -- -l ./linter_lints` in the root of `rust-linting`

DevinR528:
It seems like it fails because one of the commands expects the linter_api build.rs binary/executable file to be in `LINT_KRATES_OUT_DIR` but it's still in the normal `target/debug/build` instead of `target/linter/build/debug/build`

xFrednet:
Hmm, that's weird. The script only sets the build version, that should be independent. Maybe `cargo_linter` sets an environment value that it shouldn't :thinking: Could you go a bit more into detail how you found this out. It didn't happen on my end, and therefore it's hard to fix :sweat_smile:

DevinR528:
Yeah, these are the worst kinda bugs to track down! This only happens after the `Compiling Driver:`, `Compiling Lints` and 2 `Registering Lints` println's it then prints an extern item from (what I think) is the linter_api's build script or it could be indexmap/libc maybe. Then it fails saying that the build script for the linter_api wasn't found in `target/debug/build/linter-api-bighash/build-script-build`.

xFrednet:
> May also be relevant: I removed all the prints in the lint except extern items and functions. (relevant in that there were many less prints so maybe easier to see the error)

That could be relevant. Do you have these changes on a branch? :upside_down_face:

DevinR528:
Sorry I missed the notification for this. here it is https://github.com/rust-linting/rust-linting/pull/14

xFrednet:
[...]

PR #14, opened 2022-05-22
Fix a few typos, add my printing changes

DevinR528:
Some of the spelling fixes would be in a real PR but this is only to try and figure out #13

### DIFF
--- a/cargo-linter/src/main.rs
+++ b/cargo-linter/src/main.rs
@@ -66,7 +66,7 @@ fn main() {
     }
 
     if lint_crates.iter().any(|path| path.contains(';')) {
-        eprintln!("The absolut paths of lint crates are not allowed to contain a `;`");
+        eprintln!("The absolute paths of lint crates are not allowed to contain a `;`");
         exit(-1);
     }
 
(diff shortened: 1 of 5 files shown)


xFrednet (commented):
Thank you for the PR, I'm happy to see some action. If you have a rough code overview I can also assign my next PRs to you :)


I hope that these comments aren't too nit-picky. We can obviously still change the API during initial development. I haven't tested the change jet. I hope that I can make some progress on this tomorrow.

DevinR528:
Sorry I wasn't more clear, I just meant this to be a diff so we could hopefully reproduce my error.

I started with trying to hook into rustc's error reporting but since I couldn't get it to run without the error I just pushed what I had. Once I get the error figured out I'll use your feedback to shape a new PR!

xFrednet:
Ahh, sorry for the confusion. The title gave me the expression, that this was an actual PR. I'll check these changes out. :upside_down_face:

DevinR528:
Yeah, upon rereading, that's where I should have put the "ignore as PR", no worries though it's not like I won't use the feedback. If we could hold off on closing this PR so I can easily check it that'd be cool.

PR #15, opened 2022-05-24
Allow binary creation in rustc driver for build scripts

### LINKED ISSUE #13: Fails to build... but still kinda works

xFrednet:
Closes #13 

See [this comment](https://github.com/rust-linting/rust-linting/issues/13#issuecomment-1135583927) for an explanation.

r? @DevinR528

### DIFF
--- a/linter_driver_rustc/src/main.rs
+++ b/linter_driver_rustc/src/main.rs
@@ -44,14 +44,6 @@ impl rustc_driver::Callbacks for LinterCallback {
             lint_store.register_late_pass(register);
         }));
     }
-
-    fn after_analysis<'tcx>(
-        &mut self,
-        _compiler: &rustc_interface::interface::Compiler,
-        _queries: &'tcx rustc_interface::Queries<'tcx>,
-    ) -> rustc_driver::Compilation {
-        rustc_driver::Compilation::Stop
-    }
 }
 
 fn register() -> Box<


linter_driver_rustc/src/main.rs
```diff
@@ -44,14 +44,6 @@ impl rustc_driver::Callbacks for LinterCallback {
             lint_store.register_late_pass(register);
         }));
     }
-
-    fn after_analysis<'tcx>(
-        &mut self,
-        _compiler: &rustc_interface::interface::Compiler,
-        _queries: &'tcx rustc_interface::Queries<'tcx>,
-    ) -> rustc_driver::Compilation {
-        rustc_driver::Compilation::Stop
```
DevinR528, line 53: Wow, I would never have guessed that this would have been the culprit; what delicate things compilers are, haha.

DevinR528 (approved):
It works!

xFrednet:
Perfect!

linter_driver_rustc/src/main.rs
xFrednet: True, never would have guessed that this would cause problems, when I added it :sweat_smile:

Issue #6, opened 2022-04-24
Remove traits from stable interface


xFrednet:
The public interface currently contains some traits like `Span<'ast>`, `Ty<'ast>` and `*Item`. This is cool to implement the functions in the driver. However, this exposure of traits enables users to also implement the traits themselves and pass them back to the context. It also requires the drivers to implement other required traits, like `Debug`, `Hash` etc.

Instead, it would be nice to wrap all these into structs which will be passed to _lint crates_. In that way, we can also have a uniform driver independent `Debug` and `Display` implementation.

xFrednet:
In #11 I've tried to use a wrapper type in `linter_api` and a type alias. The result is also documented as part of that PR. I'm not convinced that a wrapper type with an alias is the best solution. Still worth a try :upside_down_face:

xFrednet:
In #11 and #16, I've tested different non-trait-based representations. For the evaluation, I've also provided screenshots of the documentation, as I think that this is a crucial aspect when it comes to the design of our user-facing API types. The examples are documented in the specific PRs. For completion, I wanted to provide a screenshot of the documentation, when traits are used:

![image](https://user-images.githubusercontent.com/17087237/170476511-5724b390-a894-4b45-9e31-4ee19e668e3d.png)

xFrednet:
A note, since I've now again added a few traits to the API. Traits themself do have problems, and their usage will need to be evaluated. This issue was mainly about trait objects (`dyn Trait` or `impl Trait`). These have now been removed from the API

PR #16, opened 2022-05-25
Represent the static item as a struct and represent common data with helper macros

xFrednet:
This PR is related to #6. Basically, I want to remove trait objects from the interface, but also don't copy the same data in each node over and over again. This PR uses a macro to do the common data implementation for each node. Common data for a specific item can now be implemented as follows:

```rs
#[derive(Debug)]
pub struct SomeItem<'ast> {
    data: CommonItemData<'ast>,
    // Node specific data
}

// This provids access to data stored in 
super::impl_item_data!(StaticItem, Static);

// Now just add functions for node specific data
impl<'ast> SomeItem<'ast> {}
```

I also like how the documentation for the created structs looks. The red border marks the node-specific functions:

image

This is a huge improvement to the version of #11 IMO

---

This implementation heavily depends on the representation used for nodes. See https://github.com/rust-linting/design/issues/13

---

This also adds a dependency to [visibility](https://crates.io/crates/visibility) which should be fine.

### DIFF
--- a/linter_api/src/ast/item.rs
+++ b/linter_api/src/ast/item.rs
@@ -474,7 +474,6 @@ pub enum Visibility<'ast> {
     /// Visible in the current module, equivialent to `pub(in self)` or no visibility
     PubSelf,
     PubCrate,
-    /// FIXME: Add a path value to this
     PubPath(&'ast Path<'ast>),
     PubSuper,
 }
--- a/linter_api/src/ast/item.rs
+++ b/linter_api/src/ast/item.rs
@@ -126,7 +126,7 @@ macro_rules! impl_item_data {
             }
 
             fn as_item(&'ast self) -> crate::ast::item::ItemType<'ast> {
-                crate::ast::item::ItemType::$enum_name(self)
+                $crate::ast::item::ItemType::$enum_name(self)
             }
 
             fn get_attrs(&self) {
--- a/linter_api/src/ast/item.rs
+++ b/linter_api/src/ast/item.rs
@@ -46,9 +46,7 @@ pub trait ItemData<'ast>: Debug {
     /// This returns this [`ItemData`] instance as a [`ItemType`]. This can be usefull for
     /// functions that take [`ItemType`] as a parameter. For general function calls it's better
     /// to call them directoly on the item, instead of converting it to a [`ItemType`] first.
-    fn as_item(&'ast self) -> ItemType<'ast> {
-        todo!()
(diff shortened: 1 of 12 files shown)


xFrednet:
@DevinR528 would you mind reviewing this? And what do you think of this change?

linter_driver_rustc/Cargo.toml
```diff
@@ -11,3 +11,6 @@ linter_api = { path = "../linter_api", features = ["driver-api"] }
 linter_adapter = { path = "../linter_adapter" }
 
 bumpalo = "3.9.1"
+
+[package.metadata.rust-analyzer]
+rustc_private=true
```
DevinR528, line 16: Did this fix it, that's funny; I had the same problem where rust-analyzer quit giving me IntelliSense for rustc and tried the same fix but couldn't get it working.

linter_api/src/ast/item.rs
```diff
@@ -100,41 +98,50 @@ macro_rules! impl_item_type_fn {
...
-    }
+            fn get_name(&self) -> Option<crate::ast::Symbol> {
+                self.data.name
+            }
 
-    fn get_name(&self) -> Option<Symbol> {
-        self.name
-    }
+            fn as_item(&'ast self) -> crate::ast::item::ItemType<'ast> {
+                crate::ast::item::ItemType::$enum_name(self)
```
DevinR528, line 129: nit:
```suggestion
                $crate::ast::item::ItemType::$enum_name(self)
```

DevinR528 (commented):
I think hiding the dyn trait makes sense, and splitting the common fields out and implementing them seems like a win. The `visibility` crate is a cool trick that will be nice for hiding things from the lint authors that the driver crate will need!

I don't see anything wrong/bad about this approach!

DevinR528 (approved):

linter_driver_rustc/Cargo.toml
xFrednet: Yeah, rs can be a bit iffy with rustc, this fixed it for me :upside_down_face:

linter_api/src/ast/item.rs
xFrednet: I always forget these, nice catch!

xFrednet:
Alright, thank you very much for your review!

I'll probably try to migrate more items from the trait to this representation, to have more examples. :upside_down_face:

PR #17, opened 2022-05-26
Use rustc's error reporting mechinery

DevinR528:
Enables using `rustc_lint::LateContext` to build warnings/errors for us from our `Lint` and `dyn Span` types.

If you want I can clean up the first few commits, so it's just, fix typos, change `Lint` as one commit, then the unsafe code, and change code in `linter_lints`?

### DIFF
--- a/linter_api/src/context.rs
+++ b/linter_api/src/context.rs
@@ -26,7 +26,6 @@ impl<'ast> AstContext<'ast> {
     }
 }
 
-
 /// This trait provides the actual implementation of [`AstContext`]. [`AstContext`] is just
 /// a wrapper type to avoid writing `dyn` for every context and to prevent users from
 /// implementing this trait.
--- a/linter_api/src/context.rs
+++ b/linter_api/src/context.rs
@@ -14,7 +14,9 @@ impl<'ast> AstContext<'ast> {
     pub fn new(cx: &'ast dyn DriverContext<'ast>) -> Self {
         Self { _cx: cx }
     }
+}
 
+impl<'ast> AstContext<'ast> {
     pub fn warn(&self, s: &str, lint: &Lint) {
         self._cx.warn(s, lint);
     }
@@ -23,6 +25,7 @@ impl<'ast> AstContext<'ast> {
         self._cx.warn_span(s, lint, sp);
     }
 }
+
 
 /// This trait provides the actual implementation of [`AstContext`]. [`AstContext`] is just
 /// a wrapper type to avoid writing `dyn` for every context and to prevent users from
--- a/linter_api/src/context.rs
+++ b/linter_api/src/context.rs
@@ -6,23 +6,23 @@ use crate::{
 /// This context will be passed to each [`super::LintPass`] call to enable the user
 /// to emit lints and to retieve nodes by the given ids.
 pub struct AstContext<'ast> {
-    _cx: &'ast dyn DriverContext<'ast>,
+    cx: &'ast dyn DriverContext<'ast>,
 }
 
 #[cfg(feature = "driver-api")]
 impl<'ast> AstContext<'ast> {
     pub fn new(cx: &'ast dyn DriverContext<'ast>) -> Self {
-        Self { _cx: cx }
+        Self { cx }
     }
 }
 
 impl<'ast> AstContext<'ast> {
-    pub fn warn(&self, s: &str, lint: &Lint) {
-        self._cx.warn(s, lint);
+    pub fn emit_lint(&self, s: &str, lint: &Lint) {
+        self.cx.emit_lint(s, lint);
     }
 
-    pub fn warn_span(&self, s: &str, lint: &Lint, sp: &dyn Span<'_>) {
-        self._cx.warn_span(s, lint, sp);
+    pub fn emit_lint_span(&self, s: &str, lint: &Lint, sp: &dyn Span<'_>) {
+        self.cx.emit_lint_span(s, lint, sp);
     }
 }
 
@@ -30,8 +30,8 @@ impl<'ast> AstContext<'ast> {
 /// a wrapper type to avoid writing `dyn` for every context and to prevent users from
 /// implementing this trait.
 pub trait DriverContext<'ast> {
-    fn warn(&self, s: &str, lint: &Lint);
-    fn warn_span(&self, s: &str, lint: &Lint, sp: &dyn Span<'_>);
+    fn emit_lint(&self, s: &str, lint: &Lint);
+    fn emit_lint_span(&self, s: &str, lint: &Lint, sp: &dyn Span<'_>);
 }
 
 /// This trait is used to create [`Symbol`]s and to turn them back into
--- a/linter_api/src/context.rs
+++ b/linter_api/src/context.rs
@@ -17,11 +17,11 @@ impl<'ast> AstContext<'ast> {
 }
 
 impl<'ast> AstContext<'ast> {
-    pub fn emit_lint(&self, s: &str, lint: &Lint) {
+    pub fn emit_lint(&self, s: &str, lint: &'ast Lint) {
         self.cx.emit_lint(s, lint);
     }
 
-    pub fn emit_lint_span(&self, s: &str, lint: &Lint, sp: &dyn Span<'_>) {
+    pub fn emit_lint_span(&self, s: &str, lint: &'ast Lint, sp: &dyn Span<'_>) {
         self.cx.emit_lint_span(s, lint, sp);
     }
 }
@@ -30,8 +30,8 @@ impl<'ast> AstContext<'ast> {
 /// a wrapper type to avoid writing `dyn` for every context and to prevent users from
 /// implementing this trait.
 pub trait DriverContext<'ast> {
-    fn emit_lint(&self, s: &str, lint: &Lint);
-    fn emit_lint_span(&self, s: &str, lint: &Lint, sp: &dyn Span<'_>);
+    fn emit_lint(&self, s: &str, lint: &'ast Lint);
+    fn emit_lint_span(&self, s: &str, lint: &'ast Lint, sp: &dyn Span<'_>);
 }
 
 /// This trait is used to create [`Symbol`]s and to turn them back into
--- a/linter_api/src/context.rs
+++ b/linter_api/src/context.rs
@@ -1,4 +1,7 @@
-use crate::ast::Symbol;
+use crate::{
+    ast::{Span, Symbol},
+    lint::Lint,
+};
 
 /// This context will be passed to each [`super::LintPass`] call to enable the user
 /// to emit lints and to retieve nodes by the given ids.
@@ -11,12 +14,23 @@ impl<'ast> AstContext<'ast> {
     pub fn new(cx: &'ast dyn DriverContext<'ast>) -> Self {
         Self { _cx: cx }
     }
+
+    pub fn warn(&self, s: &str, lint: &Lint) {
+        self._cx.warn(s, lint);
+    }
+
+    pub fn warn_span(&self, s: &str, lint: &Lint, sp: &dyn Span<'_>) {
+        self._cx.warn_span(s, lint, sp);
+    }
 }
 
 /// This trait provides the actual implementation of [`AstContext`]. [`AstContext`] is just
 /// a wrapper type to avoid writing `dyn` for every context and to prevent users from
 /// implementing this trait.
-pub trait DriverContext<'ast> {}
+pub trait DriverContext<'ast> {
+    fn warn(&self, s: &str, lint: &Lint);
+    fn warn_span(&self, s: &str, lint: &Lint, sp: &dyn Span<'_>);
+}
 
 /// This trait is used to create [`Symbol`]s and to turn them back into
 /// strings. It might be better to have a single struct like rustc does
--- a/linter_driver_rustc/src/ast/rustc.rs
+++ b/linter_driver_rustc/src/ast/rustc.rs
@@ -1,4 +1,4 @@
-use rustc_lint::{LintStore, LateContext};
+use rustc_lint::{LateContext, Level as RustcLevel, Lint as RustcLint, LintContext, LintStore};
 use rustc_middle::ty::TyCtxt;
 
 use linter_api::{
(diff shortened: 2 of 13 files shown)


xFrednet:
The last PR created some conflics, which should be simple to fix. Could you rebase?

Currently I'm debating different ways to implement lint crate tests. When done this should enable us to test out lints from `linter_lints` and maybe have lint creates in out test folder as well :)

I'll still try to get to a review today, as the conflics are simple enough to not effect the overall implementation :)

linter_driver_rustc/src/ast/rustc.rs
```diff
@@ -12,11 +11,8 @@ use linter_api::{
 
 use super::{ty::RustcTy, RustcLifetime, RustcSpan};
 
-#[expect(unused)]
 pub struct RustcContext<'ast, 'tcx> {
     pub(crate) lint_ctx: &'ast LateContext<'tcx>,
-    pub(crate) tcx: TyCtxt<'tcx>,
-    pub(crate) lint_store: &'tcx LintStore,
```
DevinR528, line 17: I made this a single commit because I'm not sure it's better to remove the fields. `LateContext` has the `LintStore` and `TyCtxt` so it seems a bit redundant? We need the `LateContext` if we want to use the warn/error reporting of rustc.

xFrednet: I remember not using `LateContext` due to some lifetime problems, as it doesn't live for `'tcx` but this might no longer be important. Even without it we could span a lint with [`struct_span_lint_hir`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.struct_span_lint_hir). That is also the function used by `LateContext` internally. But if everything works like this, it should be fine :)

xFrednet:
> Example output so you don't have to run it yourself
> 
> ```
> warning: hey there is a static item here
>   --> linter_api/src/lib.rs:10:1
>    |
> 10 | pub static LINTER_API_VERSION: &str = env!("CARGO_PKG_VERSION");
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>    |
>    = note: `#[warn(test_lint)]` on by default
> ```

Nice! I'm currently thinking of ways to implement simple UI tests for the user, but also for us. Then we can hopefully also have tests like rustc and Clippy uses :)

linter_driver_rustc/src/ast/rustc.rs
```diff
@@ -26,7 +24,61 @@ impl<'ast, 'tcx> std::fmt::Debug for RustcContext<'ast, 'tcx> {
...
+        // think we have to. We aren't returning a `dyn Trait + 'static` like if we were actually using the
+        // `dyn Any` or the `dyn Span<'_>`. We can't use the `Any::downcast_ref` machinery here since we
+        // have a non `'static` trait object in `Span` (at least I couldn't get it to work)
+        let down_span: &RustcSpan<'ast, 'tcx> = unsafe {
+            let sp_ptr = sp as *const _ as *const (*mut (), *mut ());
+            &*(sp_ptr as *const dyn std::any::Any as *const _)
+        };
+
+        self.lint_ctx
+            .struct_span_lint(to_leaked_rustc_lint(lint), down_span.span, |diag| {
```
xFrednet, line 76: This `to_leaked_rustc_lint()` call means that every time this function is called, a new lint instance is created and leaked. This is a bit problematic IMO. Another problem is, that rustc internally uses the memory address of lints for comparison, since these are static.

Ideally, we would create these lint instances in [linter_driver_rustc/src/main.rs](https://github.com/rust-linting/rust-linting/blob/1ed2c0a96ef282bf65f568bd18fc43d3faf039ab/linter_driver_rustc/src/main.rs#L40-L45) so that rustc can track the lint level. The issue with this is, that we would then have to somehow store all created lints and later save them in `RustcContext` and use them for emission. This in turn requires the lint modules to be loaded at that point as well. I've tried to do that, but ran into lifetime problems...

As an intermediate solution, I think that we should add a `HashMap` to `RustcContext` that can map from our lint instances to Rustc's leaked lint instances. For the emission, we would then only create a new instance if it's the first lint trigger and otherwise reuse the already create one.

Does this make sense? :upside_down_face:

linter_driver_rustc/src/ast/rustc.rs
```diff
@@ -26,7 +24,61 @@ impl<'ast, 'tcx> std::fmt::Debug for RustcContext<'ast, 'tcx> {
     }
 }
 
-impl<'ast, 'tcx> DriverContext<'ast> for RustcContext<'ast, 'tcx> {}
+fn to_leaked_rustc_lint(lint: &Lint) -> &'static RustcLint {
+    Box::leak(Box::new(RustcLint {
+        name: lint.name,
```
xFrednet, line 29: Another tiny nit: The name should ideally have some kind of prefix. Later, I would like something that indicates the lint crate and maybe that it comes from our tool. For now I think that `linter` would be a good prefix. Then the emission would say `` `#[warn(linter::test_lint)]` on by default``

xFrednet (commented):
I ran it locally and like the output. I've added some comments, hopefully not too nit-picky. For the difficult ones, I've tried to also add a simple workaround solution to not hold this PR up :)

DevinR528:
It might take a day or 2 for me to get to this (just moved down to Texas for a summer internship) but I agree with all the feedback (also not sure why I didn't fix some of the stuff it seems like was a repeat of my earlier draft PR, sorry about that, I guess copy/pasting is never as great as you think :laughing:). 

I had one thing that I noticed while working on this, how do you feel about the name for `lint_driver_rustc::ast::RustcWhatever` (replace "Whatever" with the name of an item type i.e. Item, Ty, etc)? I can't think of a better one. They are the concrete types for the traits so maybe `ConcreteWhatever` or `BackingWhatever` (neither very good) :shrug: just a thought, I initially thought they were renames of the actual rustc types `use rustc_middle::Item as RustcItem;` kinda thing.

xFrednet:
> It might take a day or 2 for me to get to this (just moved down to Texas for a summer internship)

That's totally fine. I hope the internship will be fun and educational :).

> I guess copy/pasting is never as great as you think :laughing:).

Good old copy/pasta :spaghetti: :joy: 

> I had one thing that I noticed while working on this, how do you feel about the name for `lint_driver_rustc::ast::RustcWhatever` (replace "Whatever" with the name of an item type i.e. Item, Ty, etc)? I can't think of a better one. They are the concrete types for the traits so maybe `ConcreteWhatever` or `BackingWhatever` (neither very good) shrug just a thought, I initially thought they were renames of the actual rustc types `use rustc_middle::Item as RustcItem;` kinda thing.

For most items in `linter_api::ast::` I actually want to move away from traits. For reasons documented in #6. With that, we'll have less direct structs in `lint_driver_rustc::ast::` than we have right now. But we'll still have some, for instance for the context. My initial idea with this naming convention, was to provide us with information, which driver created an item. I see your point, that these at first give the impression of just been an alias. By now I also think that it's unimportant to store the driver, since there will only be one active at the time, and right now, we only have rustc.

Maybe we could go the Java/C# route and call them `WhateverImpl`. I've seen some instances, where this didn't fit in java, but it worked well enough. Another Option could be `DriverWhatever`, for me, it's also fine to keep `RustcWhatever`.

What are your thoughts on the matter? (We might want to move this to a separate issue, to keep track of this)

linter_driver_rustc/src/ast/rustc.rs
```diff
@@ -1,20 +1,18 @@
...
+    lint::{Level, Lint, MacroReport},
 };
 
 use super::{ty::RustcTy, RustcLifetime, RustcSpan};
 
-#[expect(unused)]
 pub struct RustcContext<'ast, 'tcx> {
-    pub(crate) tcx: TyCtxt<'tcx>,
-    pub(crate) lint_store: &'tcx LintStore,
+    pub(crate) lint_ctx: &'ast LateContext<'tcx>,
```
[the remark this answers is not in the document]
DevinR528: Perfect, I like it!

linter_api/src/context.rs
```diff
@@ -13,10 +16,23 @@ impl<'ast> AstContext<'ast> {
...
+    }
+}
+
 /// This trait provides the actual implementation of [`AstContext`]. [`AstContext`] is just
 /// a wrapper type to avoid writing `dyn` for every context and to prevent users from
 /// implementing this trait.
-pub trait DriverContext<'ast> {}
+pub trait DriverContext<'ast> {
+    fn warn(&self, s: &str, lint: &Lint);
+    fn warn_span(&self, s: &str, lint: &Lint, sp: &dyn Span<'_>);
```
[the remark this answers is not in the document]
DevinR528: Like just have an easier way to let the user emit a lint (easier compared to rustc now, I feel like it was difficult to discover the way to do this using the rustic API for the first time) so I'm with you for docs and discovery a super easy aliased or just a pass-through method kinda thing would be nice.

```rust
trait DriverContext { 
    fn warn(&self, msg: &str) {
        /* Somehow access the static Lint from the user  and fill in the rest for them*/
        self.emit_lint(msg, get_static_lint_and_convert()); 
    }
}
```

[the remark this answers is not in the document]
DevinR528: ps I changed the names to `emit_lint[*]`

linter_driver_rustc/src/ast/rustc.rs
DevinR528: I think the way clippy does this is like
`name: concat!("linter::", stringify!($NAME)),`
for the line in the macro right?

linter_api/src/context.rs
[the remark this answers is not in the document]
xFrednet: I feel like documentation and lint development tools, will be a major part of this project. I'm trying to already provide some helpful doc comments, but it's hard to know how everything will look like in the end. Thank you for the rename, let's keep this in mind for later :+1:

linter_driver_rustc/src/ast/rustc.rs
xFrednet: Yes :+1:, rustc also does that in the tool lint macro :upside_down_face:

linter_driver_rustc/src/ast/rustc.rs
DevinR528: Well I had to also wrap the FxHashMap in a RefCell since we call it from a method that is `&self` which I think we should keep it that way. Eventually maybe this will be `Arc<Mutex<...>>` and we can run the lints in parallel and just merge the outputs nicely somehow keeping track of ordering :shrug:

anyways I'm pretty sure this is what you meant, good catch this would have been a LOT of leaking oops haha.

linter_driver_rustc/src/ast/rustc.rs
```diff
@@ -26,7 +24,61 @@ impl<'ast, 'tcx> std::fmt::Debug for RustcContext<'ast, 'tcx> {
...
+        // https://users.rust-lang.org/t/solved-transmute-between-trait-objects/13995/6
+        //
+        // In regard to the leak/forget, since we can keep this within the `'tcx` - `'ast` lifetime I don't
+        // think we have to. We aren't returning a `dyn Trait + 'static` like if we were actually using the
+        // `dyn Any` or the `dyn Span<'_>`. We can't use the `Any::downcast_ref` machinery here since we
+        // have a non `'static` trait object in `Span` (at least I couldn't get it to work)
+        let down_span: &RustcSpan<'ast, 'tcx> = unsafe {
+            let sp_ptr = sp as *const _ as *const (*mut (), *mut ());
+            &*(sp_ptr as *const dyn std::any::Any as *const _)
+        };
```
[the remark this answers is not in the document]
xFrednet: I have an idea for this specific case, which I'll work on, when I remove the span trait in favor of a struct (See #6). For now, this is as said fine

xFrednet:
Thank you for addressing the feedback. I'll do a full review tomorrow. If you want, you can rebase on master and add a UI test :upside_down_face:

linter_driver_rustc/src/ast/rustc.rs
```diff
@@ -26,7 +28,68 @@ impl<'ast, 'tcx> std::fmt::Debug for RustcContext<'ast, 'tcx> {
...
+            future_incompatible: None,
+            is_plugin: true,
+            feature_gate: None,
+            crate_level_only: false,
+        }))
+    })
+}
+
+impl<'ast, 'tcx> DriverContext<'ast> for RustcContext<'ast, 'tcx> {
+    fn emit_lint(&self, s: &str, lint: &'ast Lint) {
```
xFrednet, line 61: Small note about the function signature: we should think about users create diagnostics. Clippy has several util functions like `span_lint`, `span_lint_and_note` etc. This forces the user to always create the suggestion before the emission, even it the lint is allowed. This can also cause some ICEs, which is why Clippy wants to move away from this model. (There is an issue for this, even if there has been no progress AFAIK).

I like the diagnostic build approach of rustc. For now, this is a good starting point. I'll be busy over the next two weeks, but this is one thing I want to discuss or at least create an issue for in the design repo. :upside_down_face:

xFrednet (approved):
This looks good to me, I've added one informative comment. Could you rebase on master and add a UI test? Then everything should be ready. Thank you for the help :+1:

linter_driver_rustc/src/ast/rustc.rs
DevinR528: I'm not sure I understand what could be the other way of doing it. Or maybe I'm misunderstanding what you mean by suggestion and emission? An issue sounds like a good idea anyway better place to continue the discussion!

DevinR528:
Shuld be ready to go, you want me to merge this or should I wait for you in the future (once you have approved it should I feel ok to merge?)

xFrednet (approved):
Ähm, good question. Usually, if all that needs to be done is a rebase, it's okay to `r=me`. In those cases, I try to also mention that. In this cases, I wanted to see the added uitest, also to make sure that everything works as intended.

Now everything looks good, and you can r=me after another rebase. The number of conflicts should hopefully decrease with the growth of the project :upside_down_face:

DevinR528:
bors r=xFrednet

PR #18, opened 2022-05-27
Add `linter` as a cargo alias for `linter_lints`

xFrednet:
I think this makes running the lints a bit simple, while we don't have ui tests (which I want to work on next :))

r? @DevinR528


DevinR528:
I love using cargo aliases, I think this is a great idea!

DevinR528 (approved):

xFrednet:
> I love using cargo aliases, I think this is a great idea!

True, the abilities it provides are super useful :)

Issue #8, opened 2022-04-25
Change lint crate loading.


xFrednet:
The current lint crate loading implementation sends a trait object over a C ABI, this is AFAIK undefined behavior. Currently, it works quite well, but for the stable version a loading mechanism should be selected that uses defined behavior (and ideally only safe code), to ensure that the interface doesn't need to be changed in the future.

See [adapter docs](../tree/master/linter_adapter#implementation)

xFrednet:
I've done some research on the topic. The current way is definitely unstable and since the type layout can change between compilation I'm surprised that it has been working without a problem so far. ^^ I believe this issue is critical, as it might affect how we can represent information for lints.

I'll create an issue for this in the design repo, as we might get some more feedback there and do some experimenting with this.

xFrednet:
Looking into this, I feel like creating a general plugin system for Rust could be a rust-foundation grant all in itself. :sweat_smile:

jhpratt:
It certainly could be! Plugin systems are by no means easy.

xFrednet:
Yeah, I sadly see no good way, without a stable ABI. One option could be to introduce a stable ABI similar to `#[repr(C)]` which allows support for rust types. This should be fine when it's not the default. On the other hand, I'm far from an expert, and c/c++ has some problems that come from the chosen ABI so :shrug:.

xFrednet:
With the lint crate api rework of #34 and #35 I believe that we can close this issue. The linter will for now use dynamic libraries. I also heard an interesting idea to potentially add static linking for static and performance focussed implementations. This is still a good idea, but I feel like that would require a discussion in the /design repo.

Problems with the FFI interface will be tracked in a new issue :upside_down_face:

PR #20, opened 2022-05-29
0019 basic UI tests

### LINKED ISSUE #19: Add UI test support for lints
Clippy and Rustc both use ui test to test lints. These tests run the compiler on test files and compares the output to a `.stderr` file that is stored next to the save file.

* See Clippy's test directory as an example: https://github.com/rust-lang/rust-clippy/blob/master/tests/ui
* See Clippy's documentation: https://github.com/rust-lang/rust-clippy/blob/master/doc/adding_lints.md#testing
* Rustc's documentation: https://rustc-dev-guide.rust-lang.org/tests/ui.html#ui-tests

Dynlint has an extra [dylint_testing](https://crates.io/crates/dylint_testing/2.0.2) crate, which is smart and something I believe worth aiming for.

Currently, there are two major ui test crates [trybuild](https://crates.io/crates/trybuild) and [compiletest_rs](https://crates.io/crates/compiletest_rs). From these, I like the interface of `trybuild` more, as it's simpler. However, it lacks some customizability, which is required lints. Clippy therefore uses compiletest_rs. See Clippy's [`compile-test.rs` file](https://github.com/rust-lang/rust-clippy/blob/master/tests/compile-test.rs)

It's also worth noting, that both ui crates use rustc internally. This means that ui tests using these crates require us to use the rustc lint driver for tests. For now, I think this is fine, but we might want to ensure that our testing crate doesn't expose the specific driver :)

xFrednet:
This PR adds UI tests for our `linter_lints` lint crate. For now, it's just a configured `compiletest_rs`. In the future, we should probably consider creating our own test crate, like `dylint` does. See [dylint_testing](https://crates.io/crates/dylint_testing). For now, I feel like our API, way of loading lints etc is too unstable for that. Especially since we might need to rework how lint crates are loaded in the future #8. As long as we have no external users, I believe this is a good solution.

To run all UI tests you can just use `cargo uitest` in the workspace root. Specific test files can be specified with the `TESTNAME` environment value like this `TESTNAME=find_static cargo uitest`. The tests can also be blessed with the `BLESS` environment value. Updating an output file is as simple as running `BLESS=1 cargo uitest`. Both environment values can be combined. :)

@DevinR528 Would you mind reviewing this? It's also fine, if you don't have the time. I don't want to flood you with PRs. :upside_down_face: 

---

Closes #19 (For now, I will probably create an issue, to create a proper testing crate for external tests)


xFrednet:
Hey @DevinR528, some changes I plan to work on next will probably depend on this. Could you review this?

While a second pair of eyes is always better, I would also be okay with merging this as it is. Most of this will be reworked, when we create a testing framework for external linting crates (even if the setup idea might stay the same). So if you don't have the time, that's totally fine :)

DevinR528 (approved):
Sorry that took so long, working on llvm all day long then trying to have any motivation to do anything after work is hard!

Everything looks good to me! Great to have this feature from the start.

xFrednet:
Thank you very much for the review. Being motivated to do more coding can be hard. Working on LLVM sound like a lot of fun, but exhausting simultaneously.

If you currently don't have as much time, I'm also okay with merging these PRs without a proper review. Currently, I'm mostly experimenting with different representations and translating the Rust reference. I feel like this is not as critical and interesting to review.

There are some design related topics, which I'm currently looking into. I plan to create related issues in the design repo to get feedback, before working on the implementation.

DevinR528:
I don't mind reviewing as long as your fine with a few days in between. now that I am settled I feel like I'll have more energy and time in my routine to work on this and other personal projects!

xFrednet:
Happy to hear that you're settling in. In that case, I think we should for now continue like this and see how it goes. I'm also focussing on finishing some older projects to fully focus on this one.

Issue #26, opened 2022-06-30
Fix item visibility conversion for rustc


xFrednet:
With `nightly-2022-06-30` rustc's internal visibility representation changed to a simple enum that holds three vales. See https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/enum.Visibility.html

This is an interesting and perhaps a nicer representation that the one in `linter_api`. For the sync I only temporarily fixed the conversion for rustc since it was already not working and every situation.

This issue has two tasks:
1. [ ] decide an API representation for visibility (See also [Rust Reference](https://doc.rust-lang.org/reference/visibility-and-privacy.html))
2. [ ] fix the conversion in `linter_driver_rustc`

DevinR528:
This is a perfect example of what we are trying for. We need to make sure this change can be made without a semver break to the linter_api crate. Obviously it's still early enough that we are fine breaking compatibility but we should keep examples like this in mind.

xFrednet:
Exactly, that's why I've also not fixed this directly but created an issue. Choosing an representation for Visibility is a thing that will affect us for a long time. This is something I would also like to discuss in the design repo. (I just wanted to create the PR first (which failed now -.-))

xFrednet:
I'm creating an issue for this :upside_down_face:

jhpratt:
Imo having structs with everything private is probably the simplest way to be forward compatible. Methods can exist to expose what's deemed necessary to construct and use values.

jhpratt:
> I think that enums marked with `non_exhaustive` can be equivalently effective depending on the context.

Not necessarily, as the variants themselves are still always public. You might want to change the top-level structure, which would no longer be permissible. Leaving a comment there as requested momentarily.

PR #27, opened 2022-06-30
Rustup -> nightly-2022-06-30

xFrednet:
This update is sadly not super clean, but I've crated an issue #26 for the introduced bug.


xFrednet:
Clippy should now be happy

bors r+

PR #30, opened 2022-08-04
Rework export_lint_pass macro for stability and use too many macros!

xFrednet:
This PR changes the old unstable way to call external lint passes into a new one, using `extern "C"` functions to safely communicate between the `linter_adapter` and lint crates. Note that the current implementation is theoretically still unstable, since most transferred types still don't have `#[repr(C)]`. I plan to work on that step by step in the next weeks. For now, this is *stable enough* (Famous last words).

This PR adds and uses a lot of macros. By now I find them readable enough. I believe they're the best for our current purpose, as it allows us to add new lint pass functions with just a few line changes (about 5).

It would be wonderful if someone wants to review this. But these changes are a bit complex. I'll merge this if nobody has claimed this review when I would like to push my next changes (In like 1+ days)

cc: #8


xFrednet:
@bors try

Just to make sure that other platforms still work

xFrednet:
bory try

xFrednet:
bors try

xFrednet:
I'm trying to read up on some more FFI stuff, to see if this can be implemented in a simpler fashion :upside_down_face:  Until then I mark this PR as a draft. Feedback is still welcome :)

xFrednet:
I was considering passing the `dyn LintPass` over via a FFI function. That would save some boilerplate code. However, one goal listed in https://github.com/rust-linting/design is the wish to compile lint crates with stable. The driver uses rustc internals and therefore has to be compiled with a nightly version. This means that later down the line, the lint crates and driver might be compiled with a different rustc version. Therefore, it's probably the best to rely on `extern C` functions. To the user they luckily look identical. I'll add a function to check the used `linter_api` version and then this should be ready for merge :)

xFrednet:
This PR implements the information flow from the adapter to the lint crate. Currently, I'm working on the opposite way, requesting information from the driver/adapter. After some research, it looks like this can be done with function pointers. Similar to rustc's [`Providers` struct](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/query/struct.Providers.html). This is a slightly different approach from the one used here, where every function is exported on its own. For now, I would leave this approach as it is, to gather more experiences with it. The possible refactoring would only affect the `lint_pass` macro and nothing else :upside_down_face:

xFrednet:
<--

This PR reminds me of *Talking to Myself* from *sleeping with sirens* 

-->

I'm almost done with the followup changes. Therefore, I'm going to merge this now. I believe that this code will be revisited and reviewed by others anyways

bors r+

Issue #33, opened 2022-08-09
Support incompatible rustc versions


HKalbasi:
Drivers that are not rustc itself, currently need to use the compiler that are compiled with to compile user lints, not the compiler installed on the user system. Solving this probably needs a more detailed C api over the interface traits. Not that it needs to be addressed immediately, just something to consider in long-term.

HKalbasi:
Will be fixed by #30 ?

xFrednet:
The nightly version is mainly needed for the driver. In an ideal world, the driver will be moved into rust-lang/rust at one point. Even then, I believe we'll mostly use the nightly version, since the api will need to be updated with the latest addition to the language. But as you say that's far off, I can also imagine moving the driver to a different repo, if we have multiple ones.

> Will be fixed by https://github.com/rust-linting/rust-linting/pull/30 ?

No, that PR just uses a lot of undefined behavior when calling into linter-crates

HKalbasi:
Hmm, I didn't understand how this relates to nightly. I think it is possible to make some driver (not the rustc driver implemented in this repository) in stable (correct me if I'm wrong). The problem this issue mentions is, if I build my driver (say, rust-analyzer) with stable 1.60 and publish binaries for it, I can't use the stable compiler 1.70 installed on user's machine to compile their lints, and I need to download 1.60 on their system (I'm not even sure even that works. Is it ok to use two compiler with the exact same versions, but on different machines?)

xFrednet:
Ahh, okay, then I misunderstood you. Hopefully, now I got the question right:

> The problem this issue mentions is, if I build my driver (say, rust-analyzer) with stable 1.60 and publish binaries for it, I can't use the stable compiler 1.70 installed on user's machine to compile their lints, and I need to download 1.60 on their system

The answer is a bit more complicated. The driver will later call lint-crates over a C interface and ABI. Since that ABI is stable, you can combine any driver with any lint-crate iff they use the same version of the `liner_api` crate. So in theory, it would be no problem, to do this.

Now, in practice, we'll have to keep the `linter_api` crate up to date with any semantic related changes to rustc. Meaning that there will be most likely a new `linter_api` version with every rust version. So, the 1.60 driver would use `linter_api = 1.60.0`. If the lint crate uses the same version, then everything is fine. Otherwise, the driver will terminate with a warning, that the lint-crate API version doesn't match the driver version.

[...]

PR #34, opened 2022-08-20
Rework lint-crate -> driver communication 

xFrednet:
This PR reworks the *lint-crate/linter_api -> driver* communication to be ABI safe. The previous implementation was using a lot of undefined behavior. An unstable ABI is fun... though I understand why rustc does this.

The basic idea is as follows:
* Driver callbacks are provided as `extern "C"` functions which can safely be called.
* All data send over these calls is FFI safe
* Some conversion magic and FFI safety is done by `linter_api` and `linter_adapter`, which allows drivers to only implement the `DriverContext` trait

### Open TODOs

* [ ] This PR and the previous one, assume that all ast types are FFI safe which they aren't. This communication therefore is not yet totally safe. However, the concept should be fine and seems to work well enough
* [ ] Clean up the rustc driver #35 
* [ ] Rework/Expand `Span`s to support macros and other edge cases

---

@HKalbasi This PR was the main blocker for adding a new driver. Once this is merged, you should be able to test ra as another driver. I would highly recommend not looking at the current rustc driver, but implement the ra-driver based on the api/adapter crate and your own experience. In the rustc-driver I've played around with several concepts and by now it's an unmaintainable mess which I'll clean up next.

Also note that most AST nodes need to be reworked to be FFI safe. For now, I would suggest only working with the `StaticItem` node which is the newest and cleanest one. The rework of other items will probably follow that example.

If you have the time, you can also review this PR and provide feedback on the interface :upside_down_face:. If nobody as reviewed it by the next PR, I'll merge this :)

---

CC: #8 
CC: #33

### DIFF
--- a/linter_adapter/README.md
+++ b/linter_adapter/README.md
@@ -1,16 +1,21 @@
-# _The Adapter_
+# Linter Adapter 🔌
 
-The adapter connects the lint driver with lint crates. :electric_plug:
+The adapter provides a common interface for linter drivers to communicate with lint crates. It does some heavy lifting which would otherwise need to be done by each individual driver.
 
-:warning: This crate is not part of the stable API :warning: 
+⚠️ This is not part of the stable API, the adapter might change in between releases ⚠️
 
 ## Interface
 
-The adapter usually gets created from the environment. The following environment values can be used:
-* `LINTER_LINT_CRATES`: A semicolon separated list of dynamic libraries belonging to lint crates. Ideally, these should hold absolute paths.
+### Driver -> lint crate communication
 
-  Example: `/path/to/lint_crates/one.so;/path/to/lint_crates/two.so`
+The adapter can load defined lint crates and send information from the driver to all lint crates. The driver takes care of save ABI communication from the driver to the lint crates.
 
-## Implementation
+### Lint crate -> driver communication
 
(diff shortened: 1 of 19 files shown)


xFrednet:
bors try

xFrednet:
A small note regarding IDs, in this PR I've added the `SpanID` struct which left the layout up to the driver. I like this approach as it simplifies the API and gives the driver implementation more freedom. In my new dev branch, I've also reworked the other ids. (See: [`../id.rs` dev version](https://github.com/xFrednet/rust-linting/blob/035-rework-rustc-driver/linter_api/src/ast/common/id.rs)). In the driver, I then use asserts and a layout struct to access the data, like this:

```rs
#[repr(C)]
struct ItemIdLayout {
    krate: u32,
    index: u32,
}

pub fn to_api_item_id(..) -> ItemId {
    assert_eq!(size_of::<ItemId>(), size_of::<ItemIdLayout>(), "the layout is invalid");
	// ..
}
```

The size of the IDs is (as documented) not set in stone. I expect that the size of the `BodyId` will be adjusted when we have drivers that load all bodies and not just the one from the current crate. :upside_down_face:

xFrednet:
Alright, I've now reworked most of the rustc driver and have the next branch ready. I'll therefore merge this as is :upside_down_face:

bors r+

PR #36, opened 2022-08-26
Rework/Rewrite rustc driver and nightly bump

xFrednet:
This PR completely overhauls the rustc driver for the better. It throws a way the `to_api` trait and structures the conversion according to the api module structure.

As part of this, all IDs have moved to a new model, where the ID layout is left up to the driver. 

cc: https://github.com/rust-linting/rust-linting/issues/35. The main thing still missing is documentation for the driver, but I first want to work a bit more with the model until documenting it.

---

Otherwise, there is not much to say, moved to a different country and have been busy with other things.


xFrednet:
I'm merging this to create the next PR. :upside_down_face:

PR #41, opened 2022-09-15
Add `TyId` and adjust `SymbolID` and `SpanId` visibility

xFrednet:
I'm currently working on `Generics` which are tightly bound to types, traits, and items. I believe the next PR will be another monster. Therefore, I'm trying to extract some early commits like these :)


xFrednet:
It's finally time for the next PR. Types took more work than I expected and life was also a bit crazy. Still is, but that's besides the point. Let's merge this for now :)

bors r+

PR #45, opened 2022-10-21
Base type representation for semantic and syntactic types.

xFrednet:
This was... time intensive would be one word. It's actually crazy how complex Rust's types are. This covers most of it, and enough for most lint's, but still, this is pretty crazy.

Now, same procedure as the last few times. I'm happy about reviews, but I'll merge this if it has no review by the next PR :)


xFrednet:
I've been working on the rustc backend. This has lead to some minor adjustments to this representation. These adjustments will be included in the follow-up PR as this one hasn't received any reviews.

I'll merge this for now.

bors r+

xFrednet:
Tack!

Issue #53, opened 2022-11-11
Add rustc driver backend for items


xFrednet:
Follow up of #49

Issue #51, opened 2022-11-10
Add stable attribute representation


xFrednet:
See <https://doc.rust-lang.org/reference/attributes.html>

Issue #38, opened 2022-09-02
Avoid leaking internal types to linter crates


xFrednet:
The current API still exposes some enums and internal types, which should ideally only be available to drivers. This issue tracks the progress of removing/hiding them. Enums that only describe a characteristic should ideally be replaced with `is_` functions.

* [x] Hide `Mutability` from api in favor of `is_mut()`
* [x] Hide `UseKind`
* [x] Function attribute enums, like `Constness`
* [x] Mark FFI types as hidden

There are also some outdated types, which should be removed all together:

* [x] All API traits
* [x] `Sym` in favor of `SymId` and returning `&'ast str`

xFrednet:
This is also something I'm continually working on while designing the API notes. Some of these could still be ticked off by others :upside_down_face:

PR #54, opened 2022-11-11
Add items and adjust existing ones to the evolving standard

### LINKED ISSUE #6: Remove traits from stable interface

### LINKED ISSUE #49: Add stable item representation
Create an initial version of the stable item representations.

Open Items:
* [x] Module
* [x] ExternCrate
* [x] UseDeclaration
* [x] Function
* [x] TypeAlias
* [x] Struct
* [x] Enumeration
* [x] Union
* [x] ConstantItem
* [x] StaticItem
* [x] Trait
* [x] Implementation
* [x] ExternBlock

See: https://doc.rust-lang.org/reference/items.html

---

This issue only focuses on the API representation. The driver backend will be added in a separate PR. This reduces the PR size and makes sure that I focus on the representation and not the way rustc stores the data :)

xFrednet:
This was quite fun to write. It's remarkable how a good framework and setup can come together for fast and effective development :3

This closes #49, the driver implementation will be done in as part of #53. This issue also removes almost all traits from our API :tada:. The last trait will be removed with #51 this therefore also closes #6. Also, as usual, some progress on #38.

The normal disclaimer: I'm happy about reviews, but will merge this, if there has been no activity by the time I want to create the next PR. Let's hope this progress keeps up.

---

Closes #49
Closes #6 (The last will be removed as part of #51 )

CC: #53

### DIFF
--- a/linter_api/src/ast/item/use_decl_item.rs
+++ b/linter_api/src/ast/item/use_decl_item.rs
@@ -1,6 +1,6 @@
 use crate::ast::AstPath;
 
-use super::{CommonItemData, UseKind};
+use super::CommonItemData;
 
 /// A use declaration like:
 ///
@@ -29,6 +29,17 @@ pub struct UseDeclItem<'ast> {
 }
 
 super::impl_item_data!(UseDeclItem, UseDecl);
+
+#[repr(C)]
+#[non_exhaustive]
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub enum UseKind {
+    /// Single usages like `use foo::bar` a list of multiple usages like
+    /// `use foo::{bar, baz}` will be desugured to `use foo::bar; use foo::baz;`
+    Single,
+    /// A glob import like `use foo::*`
+    Glob,
+}
 
 impl<'ast> UseDeclItem<'ast> {
     /// Returns the path of this `use` item. For blob imports the `*` will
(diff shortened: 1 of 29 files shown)


linter_api/src/ast/item/use_decl_item.rs
```diff
@@ -30,6 +30,17 @@ pub struct UseDeclItem<'ast> {
 
 super::impl_item_data!(UseDeclItem, UseDecl);
 
+#[repr(C)]
+#[non_exhaustive]
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub enum UseKind {
```
xFrednet, line 36: Update `UseItem` to be like described in the reference and not how rustc handles it. This will also remove this enum

xFrednet:
I'm almost ready to create the next PR. I'll merge this one to rebase :)

bors r+

Issue #56, opened 2022-11-18
Let marker handle normal crates and cleanup some conversion code in rustc


xFrednet:
This is a collection issue of several smaller cleanup tasks that should be handled. The main goal is to have `cargo linter` run without any panics. Currently, it runs into a few `todo!()`s in the type conversion.

TODOs
* [x] Fix triggered `todo!()`s when running `cargo linter`
* [x] remove `cx` argument from ID translation functions where it's not needed
* [x] (Optionally) Use generics for functions which are currently really wordy like `to_api_item_id_from_def_id`
* [x] Add `item(ItemId)` fetch function to `AstContext` and use that in the adapter. This might be good to handle in another issue.
* [x] Store generic parameter bounds in parameters or update the documentation

xFrednet:
I'm working on this issue in [056-another-cleanup-branch](https://github.com/xFrednet/rust-linting/tree/056-another-cleanup-branch) I'll check of the completed items even if they're only on my dev branch

Issue #35, opened 2022-08-20
Rework rustc driver


xFrednet:
The rustc driver tested out several ways how an AST can be translated and how callbacks can be implemented. Now that some progress has been made on #8 we have an almost set API for how lint crates and drivers communicate.

The next step is now to restructure or rewrite the rustc driver to be maintainable and follow this new API.

Goals:
* [x] Implement a clear structure
* [ ] Document the structure
* [x] Improve maintainability
* [x] Add caching (Even if minimal)
* [x] Improve lifetime structure
	* This can still be improved but is already way better :)

xFrednet:
The structure first implemented as part of this issue has already improved the conversion process. However, after adding further nodes to the API and adding more conversion code, some limitations came up. For this reason, I'll once again try to restructure the driver code. The goal is to:
* [ ] Have consistent naming conventions (Currently, there are functions similar functions starting with different prefixes `to_`, `to_api`, `conv_`)
* [ ] Avoid passing `cx` around everywhere, I'll most likely use a single conversion struct for everything
* [ ] Define how the translation layer should usually be structure
* [ ] Cleanup lifetimes, mostly avoiding `'hir` if it's not needed
* [ ] Change location of the `Storage` struct
* [ ] Cut clear boundaries, where conversion functions should be located

And probably other things I can't think off rn. This comment is basically a new start for this issue.

PR #57, opened 2022-11-19
Add item conversion backend for rustc

### LINKED ISSUE #53: Add rustc driver backend for items

xFrednet:
This is still a rough diamond. I want to do some more cleanup, but prefer to do that in a follow-up PR, as this has again grown into a monster with ~1000 line changes. #56 lists some remaining TODOs. Some of these are also less related to items and more to the type system.

Closes #53 
Progress on #35


xFrednet:
I've updated #56 with some of the FIXME comments in this PR which can be fixed right now :)

xFrednet:
Quack quack! It probably doesn't matter what I write in these messages ^^

bors r+

PR #58, opened 2022-11-23
Add more syntactic types, update rustc impl and general clean up

### LINKED ISSUE #38: Avoid leaking internal types to linter crates

### LINKED ISSUE #56: Let marker handle normal crates and cleanup some conversion code in rustc

xFrednet:
Not a long description, as I feel like nobody is reading these. If you do, please leave me a comment, and I'll go into more detail. I'd also be very happy!!!

Reviews are welcome, but I'll merge it without, if there has been no activity by the next step.

---

Closes #56
Closes #38

### DIFF
--- a/linter_adapter/src/context.rs
+++ b/linter_adapter/src/context.rs
@@ -1,7 +1,7 @@
 use linter_api::{
-    ast::{Span, SpanOwner, SymbolId},
+    ast::{item::ItemKind, ItemId, Span, SpanOwner, SymbolId},
     context::DriverCallbacks,
-    ffi,
+    ffi::{self, FfiOption},
     lint::Lint,
 };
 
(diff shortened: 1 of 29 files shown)


xFrednet:
bors try

xFrednet:
I have the next tiny PR waiting, until I reword the cargo subcommand.

bors r+

PR #64, opened 2022-11-26
Rename everything to marker

### LINKED ISSUE #63: Update name to **marker**
I've finally decided on a name for this project. The name will be: **marker**

rust-linting was always a super descriptive, but at the same time boring name. At least for the main crates. I wanted to find a name, which works well for commands, is potentially related to Clippy and just gives the project some character.

I considered several names like `checker`, `linter`, `guppy` etc but never resonated with any of them. Well, I found `guppy` pretty cool, since I played a lot of "The binding of Isaac: Rebirth" but wasn't sure about the origin of that name and someone else already claimed that name on crates.io. So back to the drawing board and that's when I came up with the name *marker* like a text marker. `cargo marker` also roles of the tong, at least in my opinion.

So yeah, that's the name I settled on. The `marker` crate name has sadly already been claimed, but all the needed names like `cargo-marker`, `marker_api` etc not. Well, that was until 21 days ago. 
 
The next step, is now to rename all crates and name references to marker. Better to do it sooner rather than later :)

See: https://github.com/rust-linting/marker-crate-reservations

xFrednet:
I've also made a logo, but that'll have to wait until I update the documentation

Closes #63


xFrednet:
bors try

Issue #50, opened 2022-11-10
Add stable pattern representation


xFrednet:
Create an initial version of the stable pattern representation.

* Static Patterns
	* [x] IdentifierPattern
	* [x] WildcardPattern
	* [x] RestPattern
	* [x] ReferencePattern
	* [x] StructPattern
	* [x] `TupleStructPattern`: This will for now be expressed as a struct pattern. We'll have to see if this is really practical.
	* [x] TuplePattern
	* [x] SlicePattern
	* [x] OrPattern
* Patterns requiring expressions
	* [x] `LiteralPattern`
	* [x] `RangePattern`
    * [x] `PathPattern` (Requires path expressions)
* Not implemented (for now)
	* `MacroInvocation`: Macros are a whole different beast for another time :)
	* `GroupedPattern`: This is a syntactical pattern and is expressed in the AST. Therefore, we can ignore it :)

See: <https://doc.rust-lang.org/reference/patterns.html>

---

This issue only focuses on the API representation. The driver backend should be added in a separate PR. This reduces the PR size and makes sure that I focus on the representation and not the way rustc stores the data :)

xFrednet:
As usual, I'll check off the list items, when I've implemented them on my dev branch: 
* https://github.com/xFrednet/marker/tree/050-let-pattern-s

xFrednet:
The static pattern representation is done now. I'll unassign myself, until expressions have been added and the last patterns can also be implemented :)

PR #68, opened 2022-12-07
A good chunk of stable patterns for marker

xFrednet:
It's interesting to see how important patterns are to Rust. We use them all the time, often without noticing them. This adds the first few patterns, that can be expressed without expressions. Most of these can't be tested, but I'll try to add the rustc backend for them next.

CC: #50

---

As per usual, I'll merge this, if there has been no outside interaction, by the time I have the next PR ready. And reviews and feedback is very welcome!

### DIFF
--- a/marker_api/src/ast/pat/ident_pat.rs
+++ b/marker_api/src/ast/pat/ident_pat.rs
@@ -42,7 +42,7 @@ impl<'ast> IdentPat<'ast> {
     /// //  ^   ^^^
     /// //  |     +-- The pattern, that the variable is bound to
     /// //  |
-    /// //  +-------- The variable which is bound
+    /// //  +-------- The bound variable
     ///     _ => println!("x is most likely negative"),
     /// }
     /// ```
--- a/marker_api/src/ast/pat/ident_pat.rs
+++ b/marker_api/src/ast/pat/ident_pat.rs
@@ -1,0 +1,75 @@
+use crate::{
+    ast::{SymbolId, VarId},
+    context::with_cx,
+    ffi::FfiOption,
+};
+
+use super::{CommonPatData, PatKind};
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct IdentPat<'ast> {
+    data: CommonPatData<'ast>,
+    name: SymbolId,
+    var: VarId,
+    is_mut: bool,
+    is_ref: bool,
+    binding_pat: FfiOption<PatKind<'ast>>,
+}
+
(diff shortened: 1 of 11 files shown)


marker_api/src/ast/pat/ident_pat.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// The pattern, if the variable originates from a binding to a pattern.
+    /// ```
+    /// # let expr = 10;
+    /// match expr {
+    ///     x @ 0.. => println!("{x} is positive"),
+    /// //  ^   ^^^
+    /// //  |     +-- The pattern, that the variable is bound to
+    /// //  |
+    /// //  +-------- The variable which is bound
```
Niki4tap, line 45: `The bound variable`?

marker_api/src/ast/pat/struct_pat.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub fn fields(&self) -> &[StructFieldPat<'ast>] {
+        self.fields.get()
+    }
+
+    /// Returns `true` if the pattern is non exhaustive. The pattern might still
+    /// cover all fields, but have a rest pattern (`..`) to map for potential new
+    /// fields or if the struct has been marked as `non_exhaustive`.
+    pub fn is_non_exhaustive(&self) -> bool {
+        self.is_exhaustive
+    }
```
Niki4tap, line 32: This seems wrong? Maybe the field/function should be renamed?

xFrednet:
Nice catches, I'll update the PR later this week. Thank you for the review!

xFrednet:
Would you mind reviewing this again? Otherwise, I can also merge it at the next step as usual :)

marker_api/src/ast/pat/ident_pat.rs
```diff
@@ -42,7 +42,7 @@ impl<'ast> IdentPat<'ast> {
     /// //  ^   ^^^
     /// //  |     +-- The pattern, that the variable is bound to
     /// //  |
-    /// //  +-------- The variable which is bound
+    /// //  +-------- The variable which is the bound variable
```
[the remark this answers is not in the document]
xFrednet: Ups :joy:

Niki4tap:
Looks good! bors r+

Niki4tap:
Looks good!

bors r+

xFrednet:
I think the bors comment has to be on its own line right at the start, edits are a bit weird at times xD

Uhh magic xD. Thank you for the review <3

PR #71, opened 2022-12-31
Fixup internal docs

Niki4tap:
Not sure if that works here, but
r? @xFrednet

PR #72, opened 2023-01-03
(doc)comment nits

Niki4tap:
@xFrednet

image

Whew, that turned out way bigger than I thought it would be, feel free to take your time on this one, or just ignore until exams pass :sweat_smile: 

Even though this changes a lot of files, the changes themselves are very small (and mostly affect only (doc)comments).

I tried to touch only the (doc)comments, and not the actual code, but I think there is some weirdness with function names in `marker_api/src/ast/item/adt_item.rs`, and that *might be a little bit of a breaking change to the public API*, but I don't think we care yet?

Some of these are just my personal nits, so feel free to disagree or just not accept them.

### DIFF
--- a/marker_api/src/ast/common/callable.rs
+++ b/marker_api/src/ast/common/callable.rs
@@ -74,7 +74,7 @@ impl<'ast> Parameter<'ast> {
 
 impl<'ast> Parameter<'ast> {
     // Function items actually use patterns and not names. Patterns are not yet
-    // implemented though. A pattern should be good enough for now.
+    // implemented though. A name should be good enough for now.
     pub fn name(&self) -> Option<String> {
         self.name.get().map(|sym| with_cx(self, |cx| cx.symbol_str(*sym)))
     }
@@ -131,8 +131,8 @@ impl<'ast> CommonCallableData<'ast> {
     }
 }
 
-/// This macro automatically implements the [`Callable`] trait for structs that
-/// have a [`CallableData`] field called `callable_data`.
+/// This macro automatically implements the [`CallableData`] trait for structs that
+/// have a `callable_data` field.
 macro_rules! impl_callable_data_trait {
     ($self_ty:ty) => {
         impl<'ast> $crate::ast::common::CallableData<'ast> for $self_ty {
--- a/marker_api/src/ast/common/callable.rs
+++ b/marker_api/src/ast/common/callable.rs
@@ -7,7 +7,7 @@ use crate::{
 use super::{Abi, Span, SpanId, SymbolId};
 
 /// This trait provides information about callable items and types. Some
-/// properties might not be available for every callable object. In those
+/// properties might not be available for every callable object. In these
 /// cases the default value will be returned.
 pub trait CallableData<'ast> {
     /// Returns `true`, if this callable is `const`.
@@ -22,12 +22,12 @@ pub trait CallableData<'ast> {
 
     /// Returns `true`, if this callable is marked as `unsafe`.
     ///
-    /// Defaults to `false` if unspecified. Extern functions will
+    /// Defaults to `false` if unspecified. `extern` functions will
     /// also return `false` by default, even if they require `unsafe`
     /// by default.
     fn is_unsafe(&self) -> bool;
 
-    /// Returns `true`, if this callable is marked as extern. Bare functions
+    /// Returns `true`, if this callable is marked as `extern`. Bare functions
     /// only use the `extern` keyword to specify the ABI. These will currently
     /// still return `false` even if the keyword is present. In those cases,
     /// please refer to the ABI instead.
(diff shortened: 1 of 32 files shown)


xFrednet:
You really read through most of the code base, I'm really impressed. If you have any suggestions regarding the project structure or anything else, you're very welcome to share them. This is the first Rust project I'm structuring myself, and also the first AST design ^^

I'll wait with reviewing this until after the exams, so it should be this weekend. This is a gigantic help <3

xFrednet:
> that might be a little bit of a breaking change to the public API, but I don't think we care yet?

No, we aren't concerned with that yet. AFAIK, the project is not used by anyone yet, so we can break everything as much as we want to. I regularly do that ^^. Even with the first dev versions, so v0.0.1 once all nodes and a basic backend is added, we can still break the API, though we might want to keep a changelog at that time.

Defining a stability guarantee, will be on the list for v1.0.0. The goal is generally to have the most stable API possible, but I think we should define if and under which circumstances breaking changes would be allowed. Maybe we can also have an edition model, like rustc, using features? The awesome part, is that this is all up to us to decide, when the time comes :)

Niki4tap:
note: this PR got split into 3:
- This PR still has (doc)comment nits and other fixups
- #78 with API changes in ADT items
- #77 with leftovers from #71

sorry for the mess :)

Niki4tap:
I'll resolve the conflict in a bit, but I think it's good for review as-is.

marker_api/src/ast/common/callable.rs
```diff
@@ -22,15 +22,15 @@ pub trait CallableData<'ast> {
...
     /// also return `false` by default, even if they require `unsafe`
     /// by default.
     fn is_unsafe(&self) -> bool;
 
-    /// Returns `true`, if this callable is marked as extern. Bare functions
+    /// Returns `true`, if this callable is marked as `extern`. Bare functions
     /// only use the `extern` keyword to specify the ABI. These will currently
     /// still return `false` even if the keyword is present. In those cases,
-    /// please refer to the ABI instead.
+    /// please refer to the [`abi()`][Self::abi] instead.
```
xFrednet, line 33: Do you want to adjust this to use brackets for the link target? Seeing this, I wonder if this is worth a clippy lint, that unifies all link types ^^

marker_api/src/ast/ty.rs
```diff
@@ -97,26 +97,58 @@ pub enum TyKind<'ast> {
...
     // Sequence types
     // ================================
-    /// A tuple type like `()`, `(T, U)`
+    /// A tuple type like [`()`][prim@tuple], [`(T, U)`][prim@tuple]
     Tuple(&'ast TupleTy<'ast>),
-    /// An array with a known size like: `[T; n]`
+    /// An array with a known size like: [`[T; N]`][prim@array]
     Array(&'ast ArrayTy<'ast>),
-    /// A variable length slice like `[T]`
+    /// A variable length slice like [`[T]`][prim@slice]
```
xFrednet, line 113: Also, these doc comments. The `prim@` link is actually interesting. Probably one of these things that very few projects run into. And I'm happy to be part of one of these projects :joy:

xFrednet (approved):
Damn, I made numerous typos :sweat_smile: In the browser etc, I use LanguageTool, which works pretty well (from what I can tell) and has some context awareness, sadly it doesn't integrate too well with IDEs :/. Thank you for fixing all of these, writing the documentation for v0.0.1 will be fun :sweat_smile:

I have two tiny nits related to links, as we just had the topic in #82. Otherwise, it's good to go. You can r=xFrednet after the links have been updated :)

bors d+

marker_api/src/ast/common/callable.rs
Niki4tap: Sure, I wonder if `rustfmt` can format these

marker_api/src/ast/ty.rs
Niki4tap: Haha, yeah, glad to see that my "random looking into stdlib source" pays off. A bit sad that there isn't a shortcut like this for keywords, only for primitive types :(

Niki4tap:
> writing the documentation for v0.0.1 will be fun

I'll try to catch any further typos at the PR stage :)

marker_api/src/ast/common/callable.rs
xFrednet: IDK, rustfmt, usually avoids doc comments, as markdown formatting can be complicated. Clippy has a few lints that parse the markdown. With that, it should be possible :)

xFrednet: https://github.com/rust-lang/rust-clippy/issues/10176

Niki4tap:
Updated the links to follow one style, also added a small nit I didn't notice earlier in #82, didn't want to make a separate PR for it, hope it's ok here as a separate commit :sweat_smile:

bors r=xFrednet

xFrednet:
Adding a new commit is totally fine. Thank you!

PR #74, opened 2023-01-04
bors: run tests when no `.rs` files were changed

### LINKED ISSUE #73: Fix bors script, so it runs and tests everything even if not `.rs` files have been changed
The CI scripts have a restriction, so they only run, if `.rs` files have been changed, this is good for the normal script, but prevents bors from testing and merging on the `try` and `staging` branches. Removing the restriction should be enough to make the script work again.

Noticed in #71

Niki4tap:
Fixes #73

### DIFF
--- a/.github/workflows/rust_bors.yml
+++ b/.github/workflows/rust_bors.yml
@@ -3,12 +3,6 @@ name: Rust (bors)
 
 on:
   push:
-    # Don't run tests, when only text files were modified
-    paths-ignore:
-    - 'COPYRIGHT'
-    - 'LICENSE-*'
-    - '**.md'
-    - '**.txt'
     # Bors magic branches
     branches:
     - 'staging'


Niki4tap:
Uhhh, not sure if that's correct honestly :sweat_smile:

@xFrednet can you check?

.github/workflows/rust.yml
```diff
@@ -3,25 +3,13 @@ name: Rust
 
 on:
   push:
-    # Don't run tests, when only text files were modified
```
xFrednet, line 6: This is the script that normally runs for pull request and commit messages. This should still ignore the listed paths, as we don't need to run the rest if nothing changed in a commit. Also to be kind to the GH admins ^^

.github/workflows/rust_bors.yml
```diff
@@ -3,12 +3,6 @@ name: Rust (bors)
 
 on:
   push:
-    # Don't run tests, when only text files were modified
```
xFrednet, line 6: This looks correct, if you revert the other file it's good to go :)

xFrednet (commented):
That was fast. Small nit, but otherwise good!

.github/workflows/rust.yml
Niki4tap: Ah, thanks! Reverted

xFrednet (approved):

xFrednet:
A dream, thank you very much!!!

bors r+

PR #75, opened 2023-01-05
New `check_*` `fn`s

Niki4tap:
#### This PR extends `LintPass` trait with:
- `check_fn` for `FnItem`s
- `check_struct` for `StructItem`s
- `check_enum` for `EnumItem`s
- `check_field` for every field def (struct, enum variant)
- `check_variant` for every enum variant def

Also adds implementation for all of them (and `check_const_item`) in `marker_adapter`.

And a test `foo` lint, to ensure they are working :)

### DIFF
--- a/marker_adapter/src/lib.rs
+++ b/marker_adapter/src/lib.rs
@@ -33,16 +33,40 @@ impl<'ast> Adapter<'ast> {
 
         for item in krate.items() {
             self.external_lint_crates.check_item(cx, *item);
+            self.process_item(cx, item);
         }
+    }
 
-        for item in krate.items() {
-            match item {
-                ItemKind::Mod(data) => self.external_lint_crates.check_mod(cx, data),
-                ItemKind::ExternCrate(data) => self.external_lint_crates.check_extern_crate(cx, data),
-                ItemKind::Use(data) => self.external_lint_crates.check_use_decl(cx, data),
-                ItemKind::Static(data) => self.external_lint_crates.check_static_item(cx, data),
-                _ => {},
-            }
+    fn process_item(&mut self, cx: &'ast AstContext<'ast>, item: &ItemKind<'ast>) {
+        match item {
+            ItemKind::Mod(data) => {
+                self.external_lint_crates.check_mod(cx, data);
+                for item in data.items() {
+                    self.process_item(cx, item);
+                }
+            },
+            ItemKind::ExternCrate(data) => self.external_lint_crates.check_extern_crate(cx, data),
+            ItemKind::Use(data) => self.external_lint_crates.check_use_decl(cx, data),
+            ItemKind::Static(data) => self.external_lint_crates.check_static_item(cx, data),
+            ItemKind::Const(data) => self.external_lint_crates.check_const_item(cx, data),
+            // FIXME: Function-local items are not yet processed
+            ItemKind::Fn(data) => self.external_lint_crates.check_fn(cx, data),
+            ItemKind::Struct(data) => {
+                self.external_lint_crates.check_struct(cx, data);
+                for field in data.fields() {
+                    self.external_lint_crates.check_field(cx, field);
+                }
+            },
+            ItemKind::Enum(data) => {
+                self.external_lint_crates.check_enum(cx, data);
+                for variant in data.elements() {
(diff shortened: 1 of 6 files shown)


xFrednet:
This is an interesting one. Rustc has `check_*` functions for specific ast nodes, that was actually part of what inspired the AST representation. I think one thing worth discussing is, if having a check function for every item and expression type is actually advantageous?

If I look at rustc's [`LateLintPass`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/trait.LateLintPass.html) documentation, I find it a bit overwhelming and cluttered. (Note that I'm not the fastest reader xD). Based on this, I think it could be good, to only provide very few `check_*` methods for `ItemKind`, `ExprKind`, `StmtKind` kind where the user then selects the nodes they're interested in. The code would look something like this:

```
fn check_item(cx: &AstContext<'ast>, item: ItemKind<'ast>) {
	let ItemKind::Fn(fn_item) = item else {return};

	// Linting
}

// or
fn check_item(cx: &AstContext<'ast>, item: ItemKind<'ast>) {
	if let ItemKind::Fn(fn_item) = item {
		// Linting
	}
}
``` 

Especially the latter option, sadly adds some indention and maybe complexity, but to me, it feels a bit simpler, since every item lint starts off the same way. It also feels like this would make it simpler to write general documentation.

The best way to evaluate this is probably to implement both for now and see what users more and why. Do you have any thoughts on the matter? :upside_down_face:

Niki4tap:
> If I look at rustc's [`LateLintPass`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/trait.LateLintPass.html) documentation, I find it a bit overwhelming and cluttered.

I think rustc exposes so much functions because it *might* run faster, if you specify the exact thing you want from it. AFAIK all of the functions exposed, map almost directly to the visitor.

> Especially the latter option, sadly adds some indention and maybe complexity, but to me, it feels a bit simpler, since every item lint starts off the same way. It also feels like this would make it simpler to write general documentation.

You might be right, the main goal behind this was to provide friendly mapping to familiar rust constructs to new users (`fn` -> `check_fn`), but maybe forcing them to use more generic APIs will result in better code.

> The best way to evaluate this is probably to implement both for now and see what users more and why. Do you have any thoughts on the matter?

Yeah, I think it's a bit too early to decide, maybe marker should first have some users and feedback, and then we could see if more direct APIs are worth it. For now my idea is, make user APIs as feature-full as we can, and then strip unnecessary parts away nearing a milestone (0.0.1?).

xFrednet:
> I think rustc exposes so much functions because it might run faster

I don't think so, a trait caller can't check, which functions are actually implemented by the implementor, so all of them have to be called anyways. In these cases 2x one for `ItemKind` and one for the specific item.

> AFAIK all of the functions exposed, map almost directly to the visitor.

Yes, once we have a visitor implementation, I'd switch the adapter to also use that to visit all nodes :).

We should also provide a visitor tool at one point, either in the main crate or `marker_util`. That's also one reason why I consider not having them in the main lint pass, as users can then make their own very detailed visitor if they need to.

> For now my idea is, make user APIs as feature-full as we can, and then strip unnecessary parts away nearing a milestone (0.0.1?).

The problem is, that all features require development time. In https://github.com/rust-marker/design/issues a few users have submitted user stories which was really useful. I planned v0.0.1 to be the first milestone version to throw at users and ask for feedback. I wouldn't yet post it on any forums, but approach the users that submitted user stories and also implement internal lints for marker. So, I'd say it's more about finding a balance, we can always add more features, but removing them will get more difficult as time goes on.

If it's simple, like adding these check functions, we can add them, but at the same time create an issue in rust-maker/design to review such decisions, before the first stable release. Does that sound like a plan?

Niki4tap:
Yep, that sounds right, I'm on board :)

xFrednet:
Alright, would you mind creating an issue in the design repo for this decision? I'll add this PR to my review queue :)

Niki4tap:
Done it https://github.com/rust-marker/design/issues/32

xFrednet (approved):

xFrednet:
Looks good to me, thank you! I think `marker_lints/src/lib.rs` slowly shows how different aspects of the project slowly come together, which is super exciting. 

bors r+

PR #66, opened 2022-12-01
`cargo-marker` refactoring and review

xFrednet:
I've refactored parts of cargo-marker and reviewed the implementation. I found several limitations which are now documented in #60. This PR already fixes a few of these.

CC: #60

---

Same as always: If I don't get a review, I'll merge this, when a depending PR is ready. :)

PR #80, opened 2023-01-06
Fix ICE for `-> impl Trait`

xFrednet:
This fixes an ICE for `-> impl Trait` representations. It doesn't work with `-> impl Trait<T>` yet, where `T` is a generic of the item.

I'm not a 100% happy with how the rustc backend is structured, as it feels messy and hard to follow. I'm sadly not sure how to structure it better, as I've never worked on a translation layer like this in rust. My guess is that these things are always a bit messy, but I still hope to find a cleaner solution. For now, I hope that this is good enough.

I also found a small bug with the refactoring from #66 that `cargo dogfood` didn't recompile the driver. It should now do that again :)

r? @Niki4tap

### DIFF
--- a/cargo-marker/src/driver.rs
+++ b/cargo-marker/src/driver.rs
@@ -41,7 +41,7 @@ pub fn print_driver_version() {
 }
 
 /// This tries to install the rustc driver specified in [`DEFAULT_DRIVER_INFO`].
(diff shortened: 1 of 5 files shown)


xFrednet:
bors delegate=Niki4tap

PR #83, opened 2023-01-07
Minor cleanup around the project

xFrednet:
This is a collection of smaller things which have been at the back of my mind. Each commit addresses on TODO. For the ident commit, it's noteworthy that I only changed usages in the items module.

It's probably best to review this on a commit by commit basis.


xFrednet:
This'll probably cause some conflicts with #72. I'll try to review that PR and merge it before this'll be merged :)

xFrednet:
This should be ready for review. @Niki4tap Would you like to review it? :)

bors d=Niki4tap

marker_api/src/ast/common/span.rs
```diff
@@ -147,3 +158,73 @@ impl From<SpanId> for SpanOwner {
...
+// ```
+// impl<'ast, T> PartialEq<Ident<'_>> for T
+// where
+//     T: PartialEq<str>,
+// {
+//     fn eq(&self, other: &Ident<'_>) -> bool {
+//         self.eq(other.name())
+//     }
+// }
+// ```
```
Niki4tap, line 230:
```suggestion
macro_rules! impl_ident_eq_for {
    ($ty:ty) => {
        impl<'ast> PartialEq<$ty> for Ident<'ast> {
            fn eq(&self, other: &$ty) -> bool {
                self.name().eq(other)
            }
        }

        impl<'ast> PartialEq<Ident<'ast>> for $ty {
            fn eq(&self, other: &Ident<'ast>) -> bool {
                other.name().eq(self)
            }
        }
    };
    ($($ty:ty),+) => {
        $(
            impl_ident_eq_for!($ty);
        )+
    };
}

use impl_ident_eq_for;

impl_ident_eq_for![
    str,
    String,
    std::ffi::OsStr,
    std::ffi::OsString,
    std::borrow::Cow<'_, str>
];
```
I think that should work :)

marker_api/src/ast/common/span.rs
```diff
@@ -110,6 +110,17 @@ impl<'ast> Span<'ast> {
     }
 }
 
+impl<'ast> std::fmt::Debug for Span<'ast> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Span")
+            .field("source", &self.source)
+            .field("start", &self.start)
+            .field("end", &self.end)
+            .field("snipped()", &self.snippet())
```
Niki4tap, line 119: Maybe we should show only snippet in `Display` implementation, and leave the `Debug` one derived?

marker_api/src/ast/item.rs
```diff
@@ -42,7 +42,8 @@ pub trait ItemData<'ast>: Debug {
     fn visibility(&self) -> &Visibility<'ast>;
 
     /// This function can return [`None`] if the item was generated and has no real name
-    fn name(&self) -> Option<String>;
+    /// This function can return `None` if the item was generated and has no real name
+    fn ident(&self) -> Option<&Ident<'ast>>;
```
Niki4tap, line 46: Previous doc wasn't removed o.o

Also, I would leave the link on `None`

Niki4tap (changes requested):
Few comments, but I think in general this should be good to merge :)

marker_api/src/ast/common/span.rs
xFrednet: ... Why does it work so easily now? I understand how, but still, before I tested it for a few minutes, and it didn't work then -.-

marker_api/src/ast/item.rs
xFrednet: Oh yeah, I had a conflict while rebasing that part and must have missed that :)

marker_api/src/ast/common/span.rs
xFrednet: My though was, to split the traits like this:
* `Debug` provides all the information a user might want while debugging. This output should usually not reach the end user. We might have to take care to remove / convert unstable internal elements like `SpanId` and `SymbolId`
* `Display` can be used to create suggestions or user-facing messages.

Rustc has this split for types. With debug you get the type structure and debug information, while display gives you the type as it could be written in source code.

That was my initial thought. However, thinking about this specific case, it would probably make the output unusable, if the snippet is more than a line :sweat_smile:

Niki4tap:
Looks good to me now!

bors r+

xFrednet:
Thank you for the review, that was quick!

The next PR will probably take a bit, as it involves refactorings in the rustc's driver. I've also no lived marker, while studying for exams. With the next semester, it'll (try to) balance it more with the rest of my life. I'm still super motivated, so it'll likely continue in a similar fashion like it did :)

Issue #81, opened 2023-01-06
Allow the specification of lint crates in a file (local path and git)


xFrednet:
Currently, the lint crates have to be specified via console arguments. In a real world applications users most likely have a set of crates they want to run every time. Therefore, I suggest specifying lint crates in a file like Cargo does for dependencies.

There are several design decisions as part of this issue:
1. Which sources should be allowed?
    * I suggest mimicking the cargo format, as it has proven to work and users are used to it by now. This would mean supporting: git, path and version dependencies
2. Where should this information be specified, in `Cargo.toml` or a new `marker.toml` file?

---

Generally see: https://doc.rust-lang.org/cargo/commands/index.html for cargo comments and documentation

xFrednet:
@Niki4tap I created an issue for this feature and for discussions

Some information:
* `cargo install` sadly only works with binaries, which is sad since that could have handled all the source resolution magic for us
* Cargo will warn about weird headers in the `Cargo.toml` file, but metadata headers are allowed like: `[workspace.metadata.marker.lints]`

This information was gathered by @Niki4tap, thank you!

Niki4tap:
Tasks:
- [ ] Initial implementation (will be covered soon in a PR)
- [ ] Support git source fetching
- [ ] (Hard) Support registries
- [ ] Support features
- [ ] (Optional/Hard) Support lint configuration 
- [ ] (Need discussion) Use `cargo` library to discover nested configurations in workspaces.

Niki4tap:
Last point needs a bit of explanation, imagine:
`Cargo.toml`:
```toml
[workspace]
members = [
    "member_a",
    "member_b"
]

[workspace.metadata.marker.lints]
lint_a = { path = "lint_a" }
```

Do we want to give the ability for `member_a` to override the lints supplied in the virtual manifest (add new ones)?
Maybe with `member_a/Cargo.toml`:
```toml
[package]
# your usual keys...

[package.metadata.marker.lints]
# Should we allow adding more lints here?
lint_b = { path = "../lint_b" }
# Should `lint_a` still apply for `member_a`?
```

Niki4tap:
Additionally, using the `cargo` library would simplify source fetching and crate building a lot. The drawback for this, is that `cargo` library is unstable (and will never stabilize), just like rustc internals.

xFrednet:
Thank you @Niki4tap!

> (Hard) Support registries

[...]

Issue #67, opened 2022-12-01
Check test the driver and adapter handle `OsString`s correctly 


xFrednet:
Lint crates are compiled by `cargo-marker`, the path of the build binary is passed to the adapter via environment values. The loading of these libraries should ideally use `OsString` to ensure that it works with whatever interesting path a user might throw at it. This requires a bit of auditing.

`cargo-marker` has been checked as part of #66 and should be fine. A second pair of eyes would be appreciated :)

PR #84, opened 2023-01-09
cargo-marker: add initial configuration support

Niki4tap:
Tracked in #81

Adds initial configuration implementation, checking out one task :)

@xFrednet

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,9 @@ members = [
     "marker_adapter",
     "marker_api",
     "marker_driver_rustc",
-    # This is temporary, lint crates like this one will be managed with `cargo-liner`
+    # This is needed for rust-analyzer, uitests and other tools
     "marker_lints",
 ]
+
+[workspace.metadata.marker.lints]
+dogfood = { path = "marker_lints" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,4 +9,4 @@ members = [
 ]
 
 [workspace.metadata.marker.lints]
-dogfood = { path = "marker_lints" }
+marker_lints = { path = "marker_lints" }
--- a/cargo-marker/src/config.rs
+++ b/cargo-marker/src/config.rs
@@ -92,7 +92,7 @@ macro_rules! unsupported_fields {
 }
 
 impl Config {
-    pub fn get_marker_config() -> Result<Config, ConfigFetchError> {
+    fn get_raw_manifest() -> Result<String, ConfigFetchError> {
         let Ok(mut config_file) = File::open(CARGO_TOML) else {
             return Err(ConfigFetchError::FileNotFound);
         };
@@ -101,6 +101,11 @@ impl Config {
         if let Err(e) = config_file.read_to_string(&mut config_str) {
             return Err(ConfigFetchError::IoError(e));
         }
+        Ok(config_str)
+    }
+
+    pub fn get_marker_config() -> Result<Config, ConfigFetchError> {
+        let config_str = Self::get_raw_manifest()?;
 
         let cargo_config = match from_str::<Value>(&config_str) {
             Ok(v) => v,
--- a/cargo-marker/src/config.rs
+++ b/cargo-marker/src/config.rs
@@ -1,0 +1,147 @@
+use std::{
+    collections::HashMap,
+    fs::File,
+    io::{self, Read},
+};
+
+use serde::Deserialize;
+
+use toml_edit::easy::{from_str, Value};
+
+use crate::ExitStatus;
+
+const CARGO_TOML: &str = "Cargo.toml";
+
+#[derive(Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct Config {
+    lints: HashMap<String, LintDependency>,
+}
+
+#[derive(Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+#[serde(untagged)]
+pub enum LintDependency {
+    /// Version string like: `lint = "0.0.1"`
+    Simple(String),
+    /// Full dependency entry, like:
+    /// lint = { path = ".", version = "1.0.0" }
+    Full(LintDependencyEntry),
+}
+
+#[derive(Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct LintDependencyEntry {
+    path: Option<String>,
+    // TODO: Everything below is not yet supported
+    // Registry fetching:
+    version: Option<String>,
+    registry: Option<String>,
+    // Git source fetching:
+    git: Option<String>,
+    rev: Option<String>,
+    branch: Option<String>,
+    // Features:
+    features: Option<Vec<String>>,
+    optional: Option<bool>,
+    // TODO: do we want lint configuration here too?
+    // configuration: Option<Value>
+}
+
+#[derive(Debug)]
+pub enum ConfigFetchError {
+    /// `Cargo.toml` wasn't found
+    FileNotFound,
+    /// Read failed
+    IoError(io::Error),
+    /// Couldn't parse `Cargo.toml`
+    ParseError(toml_edit::de::Error),
+    /// `workspace.metadata.marker` has invalid structure
+    InvalidStructure,
+    /// `workspace.metadata.marker` doesn't exist
+    NotFound,
+}
+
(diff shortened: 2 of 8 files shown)


cargo-marker/src/main.rs
```diff
@@ -77,7 +110,16 @@ fn main() -> Result<(), ExitStatus> {
             .take_while(|s| s != CARGO_ARGS_SEPARATOR),
     );
 
+    let config = match Config::get_marker_config() {
+        Ok(v) => Some(v),
+        Err(e) => match e {
+            config::ConfigFetchError::NotFound => None,
+            _ => return Err(e.emit_and_convert()),
+        },
+    };
```
xFrednet, line 119: This should be moved after the flags and version check or maybe better move this into `choose_lint_crates` method, to only load it, if the user doesn't specify lints in the CLI. 

A reason to keep it in main would be, that later `install_driver` might want to also take the configuration, to allow overriding the driver version. But that's in the future and then we can load it here on demand as well.

Either solution is fine, but it should be after the version flag check

cargo-marker/src/config.rs
```diff
@@ -0,0 +1,147 @@
...
+            if let Some(ref $i) = $dep.$i {
+                eprintln!(concat!("warning: {} ({:?}): marker doesn't yet support `", stringify!($i), "` field"), $name, $i);
+            }
+        )+
+    }
+}
+
+impl Config {
+    pub fn get_marker_config() -> Result<Config, ConfigFetchError> {
+        let Ok(mut config_file) = File::open(CARGO_TOML) else {
```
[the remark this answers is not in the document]
Niki4tap: Oh, didn't know about cargo metadata, but I think invoking cargo would already be slower, and metadata contains a lot more than just the config, so I think parsing it ourselves is the right choice here

cargo-marker/src/config.rs
```diff
@@ -0,0 +1,147 @@
...
+        };
+
+        let Some(marker_config) = config_value.clone().try_into::<Config>().ok() else {
+            return Err(ConfigFetchError::InvalidStructure);
+        };
+
+        Ok(marker_config)
+    }
+
+    pub fn collect_paths(&self) -> Result<Vec<String>, ExitStatus> {
```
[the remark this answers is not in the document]
Niki4tap: It is safe, [toml spec](https://toml.io/en/v1.0.0) requires that...
> A TOML file must be a valid UTF-8 encoded Unicode document.

And we aren't doing any additional processing here, just parsing the config

Cargo.toml
```diff
@@ -4,6 +4,9 @@ members = [
     "marker_adapter",
     "marker_api",
     "marker_driver_rustc",
-    # This is temporary, lint crates like this one will be managed with `cargo-liner`
+    # This is needed for rust-analyzer, uitests and other tools
     "marker_lints",
 ]
+
+[workspace.metadata.marker.lints]
+dogfood = { path = "marker_lints" }
```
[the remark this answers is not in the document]
Niki4tap: Oh, good, another thing I didn't know about cargo :sweat_smile:

It looks like `cargo` allows it, if you also have `package = "marker_lints"`, which I have apparently missed here and in the implementation (whoops), I'll add the missing fields and I think I'll enforce the dependency renaming in the way `cargo` does, i.e. error if the package name isn't the same as renamed dependency, or the `package` field, if it exists.

cargo-marker/src/main.rs
Niki4tap: You're right, nice catch! Totally overlooked it for some reason.

Cargo.toml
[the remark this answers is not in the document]
xFrednet: Renaming lint crates was discussed in the design repo, I'd like to have the option, but generally enforce it like cargo does. We also need to know the binary name to pass it to the adapter for loading :)

cargo-marker/src/config.rs
```diff
@@ -0,0 +1,154 @@
...
+        let Ok(mut config_file) = File::open(CARGO_TOML) else {
+            return Err(ConfigFetchError::FileNotFound);
+        };
+
+        let mut config_str = String::new();
+        if let Err(e) = config_file.read_to_string(&mut config_str) {
+            return Err(ConfigFetchError::IoError(e));
+        }
+
+        let cargo_config = match from_str::<Value>(&config_str) {
```
xFrednet, line 105: Just a general comment to think about for the future:

We could split this function, to only the file loading part and the thing that parses the text. The advantage would be testability. So far, there are no real tests for `cargo-marker` and for development it's good enough if it just works. So, you don't have to add them as part of this PR. It's just something we should think about later down the line :)

cargo-marker/src/lints.rs
```diff
@@ -6,22 +6,51 @@ use std::{
 
 use crate::ExitStatus;
 
+pub struct LintCrateSpec<'a> {
```
xFrednet, line 9: Using an extra struct is more readable than the alternative of using a tuple. It should also make it easier to add crates via git etc. Could you add doc comments, explaining why and when package name can be `None`? Also, mentioning that we require the package name to be UTF-8 could be useful. TOML, forces this, but still worth a mention. That will make later audits a bit easier. (If that ever happens :))

cargo-marker/src/lints.rs
```diff
@@ -6,22 +6,51 @@ use std::{
...
+        build_local_lint_crate(self, target_dir, verbose)
+    }
+}
+
 /// This creates a debug build for a local crate. The path of the build library
 /// will be returned, if the operation was successful.
-pub fn build_local_lint_crate(crate_dir: &Path, target_dir: &Path, verbose: bool) -> Result<PathBuf, ExitStatus> {
-    if !crate_dir.exists() {
-        eprintln!("The given lint can't be found, searched at: `{}`", crate_dir.display());
+pub fn build_local_lint_crate(
```
xFrednet, line 31: Does this still need to be public?

Niki4tap:
Thanks again for the review!

I addressed the comments, did some refactoring, and I'm relatively happy with the code now, can you review it again please?

cargo-marker/src/lints.rs
```diff
@@ -6,22 +6,51 @@ use std::{
...
+    dir: &'a Path,
+}
+
+impl<'a> LintCrateSpec<'a> {
+    pub fn new(package_name: Option<&'a str>, dir: &'a Path) -> Self {
+        Self { package_name, dir }
+    }
+
+    /// Currently only checks for semicolons, can be extended in the future
+    pub fn validate(&self) -> bool {
```
xFrednet, line 20: I think the polarity of the return is confusing. I was surprised, that a later check `crate_entries.iter().any(LintCrateSpec::validate)` errored if any call returns true. I would have expected `true` to indicate that everything has been validated. Maybe `is_valid` would be a better name.

xFrednet (commented):
Generally, looks good to me, a few more NITs (Sorry if I'm too picky) and then this should be ready to go :)

cargo-marker/src/lints.rs
Niki4tap: > Also, mentioning that we require the package name to be UTF-8 could be useful

`str` implies that, so I feel that would be more explicit than it should be. 

> Could you add doc comments, explaining why and when package name can be `None`?

Sure, I'll add some doc comments

cargo-marker/src/lints.rs
Niki4tap: Nope

cargo-marker/src/lints.rs
Niki4tap: Oh, yeah, you're right, that totally needs to be renamed :sweat_smile:

cargo-marker/src/lints.rs
xFrednet: > str implies that, so I feel that would be more explicit than it should be.

Well, if someone else reads it, it could be good to know that this was an intentional choice. Using stings for paths and file names, is quite common even if it's not always correct. There is #67 because I didn't think about that during the first implementation. That's also why I noticed it.

cargo-marker/src/config.rs
Niki4tap: Done

xFrednet (approved):
Looks good to me. Thank you!!

bors r+

PR #86, opened 2023-01-11
Minor cleanup in rustc's driver

xFrednet:
This PR moves the `'ast` lifetime from the lint pass to the individual functions. My understanding is, that this scopes the lifetime to the actual function, like described in our docs. I find it hard to explain what the semantic difference, but my understanding is that having the lifetime on the trait would have allowed the user struct to use the lifetime as well, thus increasing the span of it.

I've extracted these commits from a different cleanup branch, to hopefully make that one smaller :). Nothing more to say, I hope everyone reading this has a wonderful day :upside_down_face: 

---

r? @Niki4tap Would you mind reviewing this


Niki4tap (approved):
Looks good to me!

bors r+

xFrednet:
Thank you for the review <3

Issue #69, opened 2022-12-07
Create rustc backend for static patterns


xFrednet:
#68 Adds a bunch of patterns to the API which require conversion code in rustc's driver. The patterns in question are:

* [ ] IdentifierPattern
* [ ] WildcardPattern
* [ ] RestPattern
* [ ] ReferencePattern
* [ ] StructPattern
* [ ] TupleStructPattern
* [ ] TuplePattern
* [ ] SlicePattern

Note that currently some of these might not be testable, as they can only occur in expressions. That's just something we'll have to live with :)

PR #91, opened 2023-01-12
Add simple rustc backend for patterns

xFrednet:
This PR adds the \`Unstable\` pattern and a basic rustc backend. Sadly, it's not yet testable, as it first requires bodies and expressions to be implemented. I still wanted to get most of this out of the way while I was still in the context of patterns. Most of this code has been copied from a different dev branch and adjusted to the recent refactorings.

Not too much more to say. I'd prefer to merge this now, but we could wait until expressions and bodies have been implemented to test this first. Though, it might take a while until we reach an expression with refutable patterns.

r? @Niki4tap ^^ So much for me saying that the next PR will take some time to implement :sweat_smile:. As always, feel free to skip it or do the review when you have time :).

Otherwise, to everyone reading this, have a wonderful day.

---

CC: #50 

Closes: #69

### DIFF
--- a/marker_driver_rustc/src/conversion/marker/pat.rs
+++ b/marker_driver_rustc/src/conversion/marker/pat.rs
@@ -106,8 +106,7 @@ impl<'ast, 'tcx> MarkerConversionContext<'ast, 'tcx> {
 
     fn new_rest_pat(&self) -> PatKind<'ast> {
         // This is a dummy span, it's dirty, but at least works for the mean time :)
-        let span = rustc_span::Span::default();
-        let data = CommonPatData::new(self.to_span_id(span));
+        let data = CommonPatData::new(self.to_span_id(rustc_span::DUMMY_SP));
         PatKind::Rest(self.alloc(|| RestPat::new(data)))
     }
 }
--- a/marker_driver_rustc/src/conversion/marker/pat.rs
+++ b/marker_driver_rustc/src/conversion/marker/pat.rs
@@ -1,0 +1,113 @@
+use marker_api::ast::pat::{
+    CommonPatData, IdentPat, OrPat, PatKind, RefPat, RestPat, SlicePat, StructFieldPat, StructPat, TuplePat,
+    UnstablePat, WildcardPat,
+};
+use rustc_hir as hir;
+
+use super::MarkerConversionContext;
+
+impl<'ast, 'tcx> MarkerConversionContext<'ast, 'tcx> {
+    #[must_use]
+    pub fn to_pat(&self, pat: &hir::Pat<'tcx>) -> PatKind<'ast> {
+        // Here we don't need to take special care for caching, as marker patterns
+        // don't have IDs and can't be requested individually. Instead patterns are
+        // stored as part of their parent expressions or items. Not needing to deal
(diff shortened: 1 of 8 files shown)


marker_driver_rustc/src/conversion/marker/pat.rs
```diff
@@ -0,0 +1,113 @@
...
+        let start = start.iter().map(|pat| self.to_pat(pat));
+        let middle = std::iter::once(ast_wild);
+        let end = end.iter().map(|pat| self.to_pat(pat));
+        let api_pats: Vec<_> = start.chain(middle).chain(end).collect();
+        self.alloc_slice_iter(api_pats.into_iter())
+    }
+
+    fn new_rest_pat(&self) -> PatKind<'ast> {
+        // This is a dummy span, it's dirty, but at least works for the mean time :)
+        let span = rustc_span::Span::default();
```
Niki4tap, line 109: I think it would be clearer to use [`DUMMY_SP`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_span/constant.DUMMY_SP.html), but maybe we should have a constant like this in marker itself so there would be no need to make a conversion on the next line.

Niki4tap (approved):
Just a small comment, overall looks good to me!

marker_driver_rustc/src/conversion/marker/pat.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        match &pat.kind {
+            hir::PatKind::Wild => PatKind::Wildcard(self.alloc(|| WildcardPat::new(data))),
+            hir::PatKind::Binding(hir::BindingAnnotation(by_ref, mutab), id, ident, pat) => {
+                PatKind::Ident(self.alloc(|| {
+                    IdentPat::new(
+                        data,
+                        self.to_symbol_id(ident.name),
+                        self.to_var_id(*id),
+                        matches!(mutab, rustc_ast::Mutability::Mut),
```
Niki4tap, line 26: Oh, also, shouldn't this be `hir::Mutability::Mut`?

marker_driver_rustc/src/conversion/marker/pat.rs
xFrednet: I'd say that using a dummy span is actually a bad smell. This case is intended to be temporary until, we have the first expression backend and can properly construct the span. It's annoying, that rustc doesn't store the span of `..`.

I'll create an issue to fix this, once this PR is merged :)

marker_driver_rustc/src/conversion/marker/pat.rs
xFrednet: Rustc's HIR reuses these enums from `rustc_ast`. Most likely to make the conversion code simpler and since there is no new information added. The matches macro should also trigger a type mismatch error otherwise, since it's just expanded into a simple `match` statement.

xFrednet:
Hey @Niki4tap, is the PR ready to be merged, since you resolved all conversations? :)

Niki4tap:
Oh right, forgot, to merge, sorry :v

bors r+

xFrednet:
No problem, thank you! =^.^=

Issue #52, opened 2022-11-10
Add stable expression representation


xFrednet:
Create an initial version of the stable expression representation.

* Ungrouped expressions
    * [x] [LiteralExpression](https://doc.rust-lang.org/reference/expressions/literal-expr.html)
    * [x] [PathExpression](https://doc.rust-lang.org/reference/expressions/path-expr.html)
    	* https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=295638462fb52346cd02afbe7bf3693c
    * [x] [OperatorExpression](https://doc.rust-lang.org/reference/expressions/operator-expr.html)
        * [x] AssignOperationExpressions
    * [x] [BlockExpression](https://doc.rust-lang.org/reference/expressions/block-expr.html)
        * [x] [AsyncBlockExpression](https://doc.rust-lang.org/reference/expressions/block-expr.html#async-blocks)
        * [x] [UnsafeBlockExpression](https://doc.rust-lang.org/reference/expressions/block-expr.html#unsafe-blocks)
    * [x] [ClosureExpression](https://doc.rust-lang.org/reference/expressions/closure-expr.html)
    * [ ] [AwaitExpression](https://doc.rust-lang.org/reference/expressions/await-expr.html) https://github.com/rust-marker/marker/issues/174
* Conditional expressions
    * [x] [IfExpression](https://doc.rust-lang.org/reference/expressions/if-expr.html#if-expressions)
    * [x] [IfLetExpression](https://doc.rust-lang.org/reference/expressions/if-expr.html#if-let-expressions)
    * [x] [MatchExpression](https://doc.rust-lang.org/reference/expressions/match-expr.html)
* Call expressions
    * [x] [CallExpression](https://doc.rust-lang.org/reference/expressions/call-expr.html)
    * [x] [MethodCallExpression](https://doc.rust-lang.org/reference/expressions/method-call-expr.html)
* Loop expressions
    * [x] [LoopExpression](https://doc.rust-lang.org/reference/expressions/loop-expr.html)
    * [x] [ContinueExpression](https://doc.rust-lang.org/reference/expressions/loop-expr.html#continue-expressions)
    * [x] [BreakExpression](https://doc.rust-lang.org/reference/expressions/loop-expr.html#break-expressions)
    * [x] [ReturnExpression](https://doc.rust-lang.org/reference/expressions/return-expr.html)
* Constructor expressions
[...]

PR #95, opened 2023-01-16
Adjust `StrLitExpr` node and add util and check methods

xFrednet:
While working on the rustc driver, I noticed some missing aspects and utils for statements, expressions, and bodies. The [grammar](https://doc.rust-lang.org/reference/expressions/literal-expr.html) was structured differently than I expected, which made me miss these at first.

I've extracted the API and adapter changes, to keep the PRs smaller overall.

---

Follow-up:  https://github.com/rust-marker/marker/pull/82

CC: https://github.com/rust-marker/marker/issues/52

r? @Niki4tap, would you like to review this? :)

### DIFF
--- a/marker_adapter/src/lib.rs
+++ b/marker_adapter/src/lib.rs
@@ -94,6 +94,9 @@ impl<'ast> Adapter<'ast> {
                 for stmt in block.stmts() {
                     self.external_lint_crates.check_stmt(cx, *stmt)
                 }
+                if let Some(expr) = block.expr() {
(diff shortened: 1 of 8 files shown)


xFrednet:
The amount of fixup and force push commits, shows that I probably shouldn't try to fix the CI in a lecture xD. Now everything should be fine (famous last words :sweat_smile: )

Niki4tap:
Looks good to me!

bors r+

xFrednet:
Thank you for the review!!! :)

PR #96, opened 2023-01-16
Start on rustc body, expr and stmt conversion

xFrednet:
This is the last PR I have queued up right now. It adds the rustc backend and allows us to check the first expressions and statements in lint crates :tada:.

---

This PR feels like the biggest milestone yet, as it adds one of the last conversion modules. Semantic types are the last major `rustc -> marker` module, I can think of. And we could tackle that one after this. For `marker -> rustc` direction, we have diagnostic objects and type checks, like and `implements_trait` method. These should also be unblocked soon. It's cool to see, and maybe a bit frightening, due to all the options.

I'm not sure what I'll work on next, it'll depend on my mood and how much time I have. Though, it'll likely be adding more expressions to the API. And that's enough of story telling ^^

---

r? @Niki4tap If you don't mind. Don't worry ~130 lines is just uitest output :)

### DIFF
--- a/marker_api/src/ast/stmt.rs
+++ b/marker_api/src/ast/stmt.rs
@@ -6,9 +6,9 @@ use super::{expr::ExprKind, item::ItemKind, pat::PatKind, ty::TyKind, Span, Span
 #[non_exhaustive]
 #[derive(Debug, Copy, Clone)]
 pub enum StmtKind<'ast> {
-    Item(&'ast ItemKind<'ast>),
+    Item(ItemKind<'ast>),
     Let(&'ast LetStmt<'ast>),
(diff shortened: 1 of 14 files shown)


Niki4tap (approved):
Just your usual small typo, but overall I like it! `marker_lints/src/lib.rs` really does show the project come together :)

xFrednet:
Finally, it's all coming together and we can implement the first checks. I'm hoping to implement path resolution soon, to allow us to implement the first proper lints. I already have some checks that I'd like to have for marker it self.

I'm r+ing it since you approved the PR. Thank you for the review

bors r=Niki4tap

PR #97, opened 2023-01-18
Clean up id transmutes in rustc driver

Niki4tap:
Pretty much it, just a simple macro to stop repeating the same layout checks for ids again and again.

@xFrednet :)


xFrednet (commented):
LGTM, always love to see PRs with net negative changes. I was considering, how some of this could be expressed as a macro. This is a nice solution. Thank you!

bors r+

xFrednet (approved):

Issue #85, opened 2023-01-10
Cleanup / Decide how to represent empty generics arguments


xFrednet:
While refactoring rustc's driver, I noticed that the API currently has two representations for an item/path without generic arguments.
1. The `GenericsArgs` struct is wrapped in an `Option` making it clear whether they're empty or not
2. The `GenericsArgs` are constructed with an empty slice inside them. 

Rustc also uses both of these representations. I'm not sure which one is simpler for normal users. We could just add a method `is_empty()` to `GenericsArgs`. Not having them wrapped in an extra option could be nice for lints, that want to check for specific generics :)

Niki4tap:
I don't think we should wrap them in an `Option`, `GenericArgs` with `is_empty()` method should be better in my opinion. Wrapping in an `Option` seems like a weird indirection, since you can represent the same state without it by just constructing it with an empty slice. Also, this doesn't matter a lot, but `Option<T>` a bit more memory to store than just `T`.

PR #99, opened 2023-01-24
Operator Expressions, precedence enum and some minor cleanup

xFrednet:
This PR adds a stable representation for most op expressions. The backend will be done in a different PR.

For the operation and precedence names, I was considering if full names like `Addition` would be better than `Add`. However, I think it's better to use the short hand, as most parts of the API do so, as well as rust's [traits](https://doc.rust-lang.org/std/ops/index.html#traits).

Also closes https://github.com/rust-marker/marker/issues/85

---

I took a few days off, and this is me slowly getting back into Marker. Expressions are a bit harder to *express* in a stable way. The next PR might again take a few days, though I hope to get a good rhythm going again.

r? @Niki4tap Would you mind reviewing this if you have time? :)

### DIFF
--- a/marker_api/src/ast/common/ast_path.rs
+++ b/marker_api/src/ast/common/ast_path.rs
@@ -6,11 +6,7 @@
 // split it up into an `ItemPath`, `GenericPath` etc. implementation.
 
 use super::SymbolId;
-use crate::{
-    ast::generic::GenericArgs,
-    context::with_cx,
-    ffi::{FfiOption, FfiSlice},
-};
+use crate::{ast::generic::GenericArgs, context::with_cx, ffi::FfiSlice};
 
 #[repr(C)]
 #[derive(Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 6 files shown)


Niki4tap (approved):
I like it, but to fully close #85 I think this needs to add `is_empty` method on `GenericArgs` and `FfiSlice`

xFrednet:
Good point, thank you for the tip!

xFrednet:
I just noticed that this or the previous PR completely killed `cargo dogfood` as it complains about all the unknown expressions. :joy: I'll slowly continue the grind :)

Niki4tap:
I guess that's partially on me too, since I've been approving these without testing. I think it was kind of expected, since dogfood dies because of unimplemented expression conversions in rustc driver, I'll push a PR to "fix" that in a moment, though it makes me wonder if we can test `cargo dogfood` in a CI or something like that

xFrednet:
No problem at all, I intentionally didn't include `cargo dogfood` in the CI, as I expected it to fail at times. (Until we have all expressions mapped). So don't worry about it, it was more an FYI and that I'll be fixed some time in the future. But If you create a PR to fix it, that would be fine as well :)

PR #102, opened 2023-01-27
fix panic in marker_driver_rustc

Niki4tap:
This PR replaces the `todo!` in `marker_driver_rustc` with a `eprintln!` and converts the `Expr` to an `Unstable` expression with most of the information stripped out.

It's not the best solution, but it will work for now.

Fixes `cargo dogfood` :)


xFrednet (approved):
Looks good to me, thank you for fixing my panic xD

bors r+

PR #105, opened 2023-01-31
Update license info

Niki4tap:
Keeping license texts in sync with other rust repos (design repo probably needs this too).

License badge isn't quite correct, since placeholder on crates.io is purely MIT licensed, if [marker-crate-reservations](https://github.com/rust-marker/marker-crate-reservations) can be dual-licensed, like all other repos, and updated on crates.io, that would be great!

Context:
- https://github.com/rust-lang/rust/commit/2a8807e889a43c6b89eb6f2736907afa87ae592f
- https://github.com/serde-rs/serde/pull/2355
- https://github.com/rust-lang/rust/pull/67734
- https://github.com/rust-lang/miri/pull/1846

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,7 +1,3 @@
-MIT License
-
-Copyright (c) 2022-2022 Rust-marker
-
 Permission is hereby granted, free of charge, to any
 person obtaining a copy of this software and associated
 documentation files (the "Software"), to deal in the
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 
 [![Bors enabled](https://bors.tech/images/badge_small.svg)](https://app.bors.tech/repositories/46214)
 ![CI Status](https://github.com/rust-marker/marker/actions/workflows/rust_bors.yml/badge.svg)
+[![License: MIT OR Apache-2.0](https://img.shields.io/crates/l/cargo_marker.svg)](#license)
 
 **An experimental linting interface for Rust. Let's make custom lints a reality!**
 
@@ -45,10 +46,7 @@ Internal structure of marker and its components is documented in [`docs/internal
 
 ## License
 
-Copyright (c) 2022-2022 Rust-marker
-
 Rust-marker is distributed under the terms of both the MIT license
 and the Apache License (Version 2.0).
 
 See [LICENSE-APACHE](./LICENSE-APACHE), [LICENSE-MIT](./LICENSE-MIT).
-
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -174,28 +174,3 @@ TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
    of your accepting any such warranty or additional liability.
 
 END OF TERMS AND CONDITIONS
-
-APPENDIX: How to apply the Apache License to your work.
-
-   To apply the Apache License to your work, attach the following
-   boilerplate notice, with the fields enclosed by brackets "[]"
-   replaced with your own identifying information. (Don't include
-   the brackets!)  The text should be enclosed in the appropriate
-   comment syntax for the file format. We also recommend that a
-   file or class name and description of purpose be included on the
-   same "printed page" as the copyright notice for easier
-   identification within third-party archives.
-
-Copyright 2014-2022 The Rust Project Developers
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-	http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.


README.md
```diff
@@ -8,6 +8,7 @@
 
 [![Bors enabled](https://bors.tech/images/badge_small.svg)](https://app.bors.tech/repositories/46214)
 ![CI Status](https://github.com/rust-marker/marker/actions/workflows/rust_bors.yml/badge.svg)
+[![License: MIT OR Apache-2.0](https://img.shields.io/crates/l/cargo_marker.svg)](#license)
```
xFrednet, line 11: This label is only showing the MIT license in the label, could you check on this?

Niki4tap:
@xFrednet I don't think there's much to review here, just requires an action from you to update the placeholder crate, I'm going to mark this as blocked for now, feel free to do that whenever you want

LICENSE-MIT
```diff
@@ -1,7 +1,3 @@
-MIT License
-
-Copyright (c) 2022-2022 Rust-marker
```
xFrednet, line 3: This feels a bit weird, especially since the [official license text](https://opensource.org/licenses/MIT) includes the Copyright line. Though I understand that "Rust Developers" as in the linked commits is not a legal entity and would in theory mean that every dev has a copy right. :thinking: I'd think that Rust-Marker would make the Rust-Marker organization the copy right holder which kind of makes sense. But looking at other licenses under rust-lang it seems like they also removed it, so I would be fine with this

README.md
```diff
@@ -43,12 +44,11 @@ This will start `cargo-marker`, load [`./marker_lints`](./marker_lints) as a *li
...
-
-Rust-marker is distributed under the terms of both the MIT license
-and the Apache License (Version 2.0).
-
-See [LICENSE-APACHE](./LICENSE-APACHE), [LICENSE-MIT](./LICENSE-MIT).
+#### License
 
+<sup>
+Licensed under either of <a href="LICENSE-APACHE">Apache License, Version
+2.0</a> or <a href="LICENSE-MIT">MIT license</a> at your option.
```
xFrednet, line 53: I don't think this change is quite right. rust, clippy, cargo and rust-analyzer all state `under the terms of both the MIT license and the Apache License (Version 2.0)` I took this text from one of those. Removing the copyright text again make sense.

README.md
```diff
@@ -43,12 +44,11 @@ This will start `cargo-marker`, load [`./marker_lints`](./marker_lints) as a *li
...
-## License
+---
 
-Copyright (c) 2022-2022 Rust-marker
-
-Rust-marker is distributed under the terms of both the MIT license
-and the Apache License (Version 2.0).
-
-See [LICENSE-APACHE](./LICENSE-APACHE), [LICENSE-MIT](./LICENSE-MIT).
+#### License
```
xFrednet, line 49: I'd prefer this to be a normal section with normal text size. It makes the file more uniform and this is an equal level as the other parts IMO. Is there a reason why you suggest this?

xFrednet (commented):
Hope my comments are not too nit picky :sweat_smile:  We should also check that the licenses are correct in the `Cargo.toml` files for crates.io

xFrednet:
> just requires an action from you to update the placeholder crate

The placeholder crates can remain, as they are right now. Old versions can't be removed/modified anyways. The new licenses will be updated on crates.io when we push the version from this repo. Thank you for pointing it out though :)

README.md
Niki4tap: Yep, yep, yep, it takes the license information from the place holder crate from marker-crate-reservations, which is MIT only, that why I asked you to update it, see the PR description, and my comment later

README.md
Niki4tap: That's... interesting, I would assume they have different meanings, this was copied from serde, since the original didn't sound quite right to me, but if everyone else does it this way, I probably shouldn't change this, thanks for pointing this out

README.md
Niki4tap: I stated the reason in the second comment, copy-pasting :sweat_smile:

Going to rollback that change right now anyways

Niki4tap:
> The placeholder crates can remain, as they are right now.

The badge takes licensing information from crates.io, not the github repo, that's why the crates need to be updated

README.md
xFrednet: Good to know, then I understand why this is blocked. Thank you!

README.md
xFrednet: Maybe we can take the text from Cargo to make sure, that this is correct. (Slightly adjusted) That would look like this:

---

## License

Marker distributed under the terms of both the MIT license and the Apache License (Version 2.0).

See [LICENSE-APACHE](LICENSE-APACHE) and [LICENSE-MIT](LICENSE-MIT) for details.

---

```
## License

Marker distributed under the terms of both the MIT license and the Apache License (Version 2.0).

See [LICENSE-APACHE](LICENSE-APACHE) and [LICENSE-MIT](LICENSE-MIT) for details.
```

Niki4tap: Oh, right, I remember now, I did this change because of the copyright notice in the readme, without it, it's practically the same, I think just removing the copyright line should be fine, what do you think?

xFrednet: I agree, just removing the copyright line sounds good to me :)

xFrednet:
> The badge takes licensing information from crates.io, not the github repo, that's why the crates need to be updated

Looks like we have another reason to look forward to that day. ^^ IDK if I can reach the goal of completing the AST representation by March, but I want to be optimistic.

xFrednet (approved):
We'll wait on the merge until the first version has been published, to ensure that the license shield is correct. If you want, you can also remove that line and `r=me`. Then we can add it back with the first version :)

Niki4tap:
I don't like merging this with a wrong badge, I think what I can do is hardcode the badge for now, and make it link to the crates later. Or we can just leave this as blocked until the license has been updated, as it's not really urgent and shouldn't conflict with anything

xFrednet:
I'm fine with either option. Even if there will be conflicts, they should be simple to resolve. I'll add this to the milestone, to merge is shortly before releasing the crates :)

Niki4tap:
Sure, I'm fine with leaving this blocked then

xFrednet:
Thank you for this PR! I'm currently updating the license information of the crates and `README.md` files as part of https://github.com/rust-marker/marker/issues/151. I'm closing this PR in favor of that issue.

Thank you, for the work you did, I highly appreciate it!

PR #108, opened 2023-02-08
`marker_adapter`: Support non-utf8 paths

### LINKED ISSUE #67: Check test the driver and adapter handle `OsString`s correctly

Niki4tap:
Makes `marker_adapter` use `OsString`s to support paths which might not be utf-8.

Fixes #67

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -309,6 +309,7 @@ dependencies = [
 name = "marker_adapter"
 version = "0.1.0"
 dependencies = [
+ "cfg-if",
  "libloading",
  "marker_api",
 ]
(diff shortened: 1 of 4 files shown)


Niki4tap:
#103 makes me a bit concerned of any unchecked windows changes now, so let's see what bors has to say first.

bors try

Niki4tap:
Hmmm, seems like github CI clippy doesn't trigger on `#cfg[windows]` blocks, that's not particularly helpful, let's see if things are working or not

bors try

xFrednet (approved):

xFrednet:
Looks good to me and I like the changes. Thank you!

bors r+

PR #109, opened 2023-02-08
Merge separate `*_lit_expr.rs` files into a single `lit_expr.rs` file

xFrednet:
I want to find a balance between keeping everything in one file, like rustc does with [hir](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_hir/hir.rs.html) or separating every item into its own file like I did for items. This PR groups all literal expressions into a single file. The idea is to group expressions based on the type, like I already did for operator expressions. Guess we'll see how this feels after a few more weeks of development :)

r? @Niki4tap I haven't changed anything and just copied it. You can review it if you want or skip it and `r=me` :)


Niki4tap (approved):
Seems like a pretty straightforward change, I think it looks more concise too.

bors r+

xFrednet:
That was quick, thank you! Let's hope progress continues like this :D

PR #110, opened 2023-02-10
Add API representation for constructors

xFrednet:
This PR adds an API representation for constructors. The rustc backend will be added in a follow-up PR to keep the changes reviewable.

---

cc: #52

r? @Niki4tap if you don't mind reviewing some more API changes :)

### DIFF
--- a/marker_api/src/ast/expr/ctor_expr.rs
+++ b/marker_api/src/ast/expr/ctor_expr.rs
@@ -1,6 +1,4 @@
-//! This module contains all expressions, which are typically used to construct
-//! or deconstruct data. A simple example is the [`ArrayExpr`] which can be
-//! used to create or destruct an array.
+//! This module contains all expressions, which are typically used to construct data.
 
 use crate::{
     ast::{AstQPath, Ident, Span, SpanId},
@@ -10,16 +8,13 @@ use crate::{
 
 use super::{CommonExprData, ExprKind, ExprPrecedence};
 
-/// An array expressions can be used to construct an array or destruct an array.
+/// An expression constructing an array.
 ///
 /// ```
 /// //            vvvvvvvvvvvv An array expression with four element expressions
 /// let array_1 = [1, 2, 3, 4];
-/// //            vvvvvv An array expression with one element and one len expression
+/// //            vvvvvv A repeat array expression with repeat and length operands
 /// let array_2 = [6; 3];
-///
-/// //  vvvvvvvvv An array expression destructing `array_2`
-/// let [a, b, c] = array_2;
 /// ```
 #[repr(C)]
 #[derive(Debug)]
@@ -62,14 +57,11 @@ impl<'ast> ArrayExpr<'ast> {
     }
 }
 
-/// A tuple expression used to construct or deconstruct a tuple.
+/// An expression used to construct a tuple.
 ///
 /// ```
 /// //          vvvvvvvvvvvv A tuple expression with four elements
 /// let slice = (1, 2, 3, 4);
-///
-/// //  vvvvvvvvvvvv A tuple expression destructing `slice`
-/// let (a, b, c, _) = slice;
 /// ```
 #[repr(C)]
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


marker_api/src/ast/expr/ctor_expr.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+impl<'ast> RangeExpr<'ast> {
+    pub fn start(&self) -> Option<ExprKind<'ast>> {
+        self.start.copy()
+    }
+
+    pub fn end(&self) -> Option<ExprKind<'ast>> {
+        self.start.copy()
+    }
```
Niki4tap, line 253:
```suggestion
    pub fn end(&self) -> Option<ExprKind<'ast>> {
        self.end.copy()
    }
```
Copy-paste :sweat_smile:

Niki4tap (changes requested):
Looks good overall, though, I have a couple of questions about the naming:
- Rust reference calls it a [`StructExpr`](https://doc.rust-lang.org/reference/expressions/struct-expr.html), though, there is quite a lot of [`Ctor`s](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/?search=Ctor) in rustc internals too, so I'm not sure what's really appropriate here
- Docs mention `destructing`/`destruct`/`deconstruct`, but I think the word you're looking for here is [`destructuring`](https://rust-lang.github.io/rfcs/2909-destructuring-assignment.html), if you could clarify that a bit, it would be really nice

marker_api/src/ast/expr/ctor_expr.rs
xFrednet: Maybe, I'm relying too much on your reviews, that I didn't double-check this. Thank you, this has been an enormous help as always!

xFrednet:
> Looks good overall, though, I have a couple of questions about the naming:
> 
> * Rust reference calls it a [`StructExpr`](https://doc.rust-lang.org/reference/expressions/struct-expr.html), though, there is quite a lot of [`Ctor`s](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/?search=Ctor) in rustc internals too, so I'm not sure what's really appropriate here

Rustc has two ways to construct objects, first the struct expression notation `Name {field: expr}` and for tuples and enum variants special functions, called Ctor. Writing `Enum::Var(expr)` is desugared as calling a function on the type `Enum` called `Var` with `expr` as an argument. So, in rustc, they have a semantic difference. Now, I have the feeling that this complicates stuff with no obvious benefit. Therefore, I want to join that all into one.

The name `StructExpr` would be inline with the reference, and was also the name I used at first. However, coming new to marker and rustc, I don't think it's obvious that the `StructExpr` is used to construct all types of objects with field. Other names I considered were:
* `AdtExpr` (For Arithmetic data type expression, even if not all things are an ADT)
* `TypeExpr` (This one sounds too generic)
* `TypeCtorExpr` (Clearer than the previous one, but then we can also just use `CtorExpr`)
* `FieldInitExpr` (Kind of weird, even if it describes the expression)

I don't know what the better name is, TBH. What do you think?

> * Docs mention `destructing`/`destruct`/`deconstruct`, but I think the word you're looking for here is [`destructuring`](https://rust-lang.github.io/rfcs/2909-destructuring-assignment.html), if you could clarify that a bit, it would be really nice

Yeah... I picked up calling this a deconstructor in c++ several years ago (There it's also called a destructor, but I just learned it wrong). In my head, that name also makes sense, even if it's wrong. I'll correct the docs!

---

Thank you! All of these are nice catches, maybe one day I'll be a better writer.

Niki4tap:
> Now, I have the feeling that this complicates stuff with no obvious benefit. Therefore, I want to join that all into one.

> I don't know what the better name is, TBH. What do you think?

Ah, okay, that makes sense, then `CtorExpr` should be totally fine.

Niki4tap (approved):
There are still some docs that reference `destructing`, buy you can `r=me` with these removed/fixed

xFrednet:
Thank you for the review! I'm currently working on the rustc backend, and it's going pretty good so far. I hope to have the next PR ready soon. :)

bors r=Niki4tap

PR #114, opened 2023-02-15
Add `MethodExpr` and rustc backend

xFrednet:
This PR adds the `ExprKind::Method` variant and supports the representation of unsafe blocks.

---

r? @Niki4tap My understanding is, that you're fine with me pinging you for PRs. If you have the time, I'd value your feedback. :)

cc: #52

### DIFF
--- a/marker_driver_rustc/src/context.rs
+++ b/marker_driver_rustc/src/context.rs
@@ -4,7 +4,7 @@ use marker_adapter::context::{DriverContext, DriverContextWrapper};
 use marker_api::{
     ast::{
         item::{Body, ItemKind},
-        BodyId, ItemId, Span, SpanOwner, SymbolId,
+        BodyId, ExprId, ItemId, Span, SpanOwner, SymbolId,
     },
     context::AstContext,
     lint::Lint,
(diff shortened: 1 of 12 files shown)


marker_driver_rustc/src/context.rs
```diff
@@ -113,4 +113,8 @@ impl<'ast, 'tcx: 'ast> DriverContext<'ast> for RustcContext<'ast, 'tcx> {
         let api_str: &'ast str = unsafe { std::mem::transmute(rustc_str) };
         api_str
     }
+
+    fn resolve_method_target(&'ast self, _id: ExprId) -> ItemId {
+        todo!()
+    }
```
xFrednet, line 119: I know how to implement this, but I want to do some slight adjustments beforehand, which will result in superficial changes. For the review, I believe better to split the implementation in two PRs.

Niki4tap (approved):
Looks good to me, but I think the new test is missing the normalization comment:
```rust
// normalize-stdout-windows: "tests/ui/" -> "$$DIR/"
```

Add it and you can `r=me` right after :)

xFrednet:
Perfect, thank you for the fast review. I noticed that this also ICEed dogfood again. :sweat_smile: I'm so looking forward, to when we can add `cargo dogfood` to the CI.

bors r=Niki4tap

PR #117, opened 2023-02-22
Rustc backend for operations

xFrednet:
I noticed that I forgot to add a backend for operation expressions... until now :)

---

r? @Niki4tap if you have the time

cc: #52


Niki4tap (approved):
A couple of nits, should be good to `r=me` after addressing

xFrednet:
Thank you for the review! I love how marker is slowly coming together. I'm still not sure if, v0.1.0 will be ready by the end of March, but I'm getting optimistic again :)

bors r=Niki4tap

PR #118, opened 2023-02-23
Add AssignExpr API represention

xFrednet:
This PR adds the assign expression to the API. The implementation was a bit complicated design wise, I created https://github.com/rust-marker/design/issues/35 to discuss other solutions and review this representation before v1.0.0.

Before reviewing this, I suggest checking this [Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=0da87ed83a1f80c61b026f54bf75d077) with *Show HIR* to see how rustc desugars some assign expressions. `#[clippy::dump]` and then *Tools > Clippy* can also always be helpful.

I also appreciate feedback on the design.

---

r? @Niki4tap Would you mind giving this a review? We can also wait a few days to see if we get any feedback on the design issue. On the other hand, we can always revert these changes :). Besides that, I'm trying to document all new nodes added to the API, we'll see how that goes ^^.

#52

### DIFF
--- a/marker_driver_rustc/src/conversion/marker/expr.rs
+++ b/marker_driver_rustc/src/conversion/marker/expr.rs
@@ -5,8 +5,10 @@ use marker_api::ast::{
         IndexExpr, IntLitExpr, IntSuffix, LetExpr, MatchArm, MatchExpr, MethodExpr, PathExpr, RangeExpr, RefExpr,
         StrLitData, StrLitExpr, TupleExpr, UnaryOpExpr, UnaryOpKind, UnstableExpr,
     },
+    pat::PatKind,
     Ident,
 };
+use rustc_hash::FxHashMap;
 use rustc_hir as hir;
 use std::str::FromStr;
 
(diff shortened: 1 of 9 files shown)


marker_driver_rustc/src/conversion/marker/expr.rs
```diff
@@ -314,4 +339,55 @@ impl<'ast, 'tcx> MarkerConverterInner<'ast, 'tcx> {
...
+    ///
+    /// [Playground]: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021
+    fn to_assign_expr_from_desugar(&self, block: &hir::Block<'tcx>) -> AssignExpr<'ast> {
+        let lhs_map: FxHashMap<_, _> = block
+            .stmts
+            .iter()
+            .skip(1)
+            .map(|stmt| {
+                if let hir::StmtKind::Expr(expr) = stmt.kind
+                    && let hir::ExprKind::Assign(ass, value, _span) = expr.kind
```
Niki4tap, line 369: This looks like a pretty unfortunate variable name :sweat_smile:

Maybe something like `assign_expr`?

Niki4tap (approved):
Just a few small nits, but looks good overall!

The way rustc desugars assignment expressions is interesting to say the least, I guess there is a reason for it to do that, however, the way they are represented in HIR seems pretty unintuitive and hard to work with for external tools like linters.

I really like the way you've worked around that, and honestly I don't see a way to implement it better or any immediate drawbacks. I think it should be good to merge, but I also agree that we should hear some feedback before fully committing to it.

marker_driver_rustc/src/conversion/marker/expr.rs
xFrednet: Ups, changing this was on my todo list and then I forgot ^^

xFrednet:
Thank you for the review :heart: 

> The way rustc desugars assignment expressions is interesting to say the least

True, I can imagine that it saves some code, as values can only be destructed with patterns. I was surprised that this allows the following code:

```rs
let mut a = 1;
[a, a] = [2, 3];
println!("{a}");
```

even if it officially mutates `a` twice in one statement.

> I really like the way you've worked around that, and honestly I don't see a way to implement it better or any immediate drawbacks. 

Thank you! Resugaring assign expressions looked scary, that's why I put it off for some time. I'm happy that I did, as I find this solution elegant and usable.

> I think it should be good to merge, but I also agree that we should hear some feedback before fully committing to it.

I'd also say we can merge this now. The design issue will remind us to review this before v1.0.0. It looks like loop expressions are the last major part of expressions, before I can move on to the next issue. (Ignoring async expressions, as they'll require more work to resugar ... I hope we can start some tests without that being finalized)

Niki4tap:
Thanks, let's merge this then!

bors r+

PR #126, opened 2023-04-20
cargo-marker: add git support

### LINKED ISSUE #87: Allow fetching lints from registries
#81 adds support to specify lint crates in `Cargo.toml` files. For now only local paths and git repositories are allowed. In the future, registries like `crates.io` should also be allowed.

This issue is currently blocked until the first version of `marker_api` has been released to crates.io, as we're otherwise not able to upload a lint crate to crates.io. It might be possible to work around this restriction by defining a git dependency. :thinking: I'll still mark it as blocked for now.

Niki4tap:
Its been a while since I've worked on this, but I've finally took the time to finish it.

This PR adds git support to marker so this now works:
```toml
[workspace.metadata.marker.lints]
marker_lints = { git = "https://github.com/rust-marker/marker" }
```

It makes use of [`cargo_fetch`](https://github.com/Niki4tap/cargo_fetch) library which I made for this specific use-case, which is essentially just a thin wrapper around `cargo` library (so the builds are now way slower :p).

I'll try to maintain the library to my best capacity, and prioritize marker's goals whenever needed.

Addition of the library also technically allows for any kind of package to be downloaded, so this also *technically* supports registries, though this PR does not implement it, because this is blocked on marker APIs being released.

---

In total this checks out... *\*looks up the issue\**... one task from #81!
But also *technically* closes  #87 whenever APIs are published.

review side note: most changes are from `Cargo.lock`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -18,10 +24,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2cb2f989d18dd141ab8ae82f64d1a8cdd37e0840f73a406896cf5e99502fab61"
 
 [[package]]
+name = "arrayvec"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
+
+[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
+
+[[package]]
+name = "base64"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 
 [[package]]
 name = "bitflags"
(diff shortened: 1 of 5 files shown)


xFrednet:
> It makes use of [cargo_fetch](https://github.com/Niki4tap/cargo_fetch) library which I made for this specific use-case

That is super cool! I'm not 100% sure if git dependencies work for crates we want to push to crates.io. But if that becomes a problem, we could move it to here or publish it. Let me know if you need any help with the maintenance!

Niki4tap:
> That is super cool! I'm not 100% sure if git dependencies work for crates we want to push to crates.io

Oh I know, that's why I've published it ;)

xFrednet (approved):
I'm honestly impressed. Thank you very much for all the work you put into this, even making a new library. This is so much more than I hoped!

I still want to do some small tests before merging it, but currently everything looks, awesome!

Btw, do you have a lint/repo, that you tested this implementation with? Otherwise, I might look into adding a repo to `rust-marker` and hacking a test together :)

Niki4tap:
> Btw, do you have a lint/repo, that you tested this implementation with? Otherwise, I might look into adding a repo to rust-marker and hacking a test together :)

Yeah some tests would be great, but yes, I've tested locally with the marker repo itself, you can just do:
```toml
[workspace.metadata.marker.lints]
marker_lints = { git = "https://github.com/rust-marker/marker" }
```

in the Cargo.toml, and it should fetch the test lints from the repo

xFrednet:
Hey, I finally found some time to test this. Sorry, for my inactivity. My semester will end in about one month and then I'll have more time, also to (hopefully) make the final push for v0.0.1.

Anyways, I see what you mean, that the compile time increased significantly. In the future, I would be in favor of having such a functionality in Cargo itself, there is already an issue for that. However, in the meantime, this is the best solution I can think of, and I'm really grateful that you created the library :heart: 

---

Now, related to testing:

I've reserved the `marker_lint_crate_test` crate name and would like to later include a fetch test for both git and crates.io in the CI. However, having a test crate for that (besides `marker_lints`) sounds a bit painful, until we have a working template and an idea of the stable API. AKA v0.0.1. I crated https://github.com/rust-marker/marker/issues/127 to create a template create we can use for testing, once we reach that point. Until then, I think we should merge this without CI tests.

I've tested the implementation locally. The git fetching works :tada: However, the normal path dependency:
```toml
[workspace.metadata.marker.lints]
marker_lints = { path = "marker_lints" }
```
is sadly broken. I get the following error when running `cargo dogfood` on this branch. (Without any further modifications)

```sh
$ c clean
# (Yes, I have an alias cargo, and I can highly recommend it xD)
$ c dogfood
   Compiling proc-macro2 v1.0.49
   [..]
   Compiling cargo_marker v0.1.0 (./marker/cargo-marker)
    Finished dev [unoptimized + debuginfo] target(s) in 59.85s
     Running `target/debug/cargo-marker --forward-rust-flags`
Compiling rustc driver
   Compiling proc-macro2 v1.0.49
   [...]
   Compiling marker_driver_rustc v0.1.0 (./marker/marker_driver_rustc)
    Finished dev [unoptimized + debuginfo] target(s) in 7.38s

Compiling Lints:
Failed fetching lint crates: invalid path url `marker_lints`
Error: LintCrateFetchFailed
```

Could you take a look at this? :) Once that is fixed, the PR is ready to be merged.

Niki4tap:
> Hey, I finally found some time to test this. Sorry, for my inactivity.

No problem, I'm not rushing, I have other things to deal with myself.

> Anyways, I see what you mean, that the compile time increased significantly. In the future, I would be in favor of having such a functionality in Cargo itself, there is already an issue for that.

Yeah the library seems like a hacky solution, and I would like to see it in the cargo too, but this works for now :p

> However, the normal path dependency is sadly broken.

Nice catch, funny enough it was a typo (`PackageSource::Path` the enum variant vs `PackageSource::path` the helper method), rust may protect you from memory errors, but not from typos I guess :smile:

Everything should be good now I think

xFrednet (approved):
Alright, this version looks good to me. Thank you!!!

bors r+

xFrednet:
Looks like my tests with merge queues broke bors :sweat_smile: 

Now it should work again

bors r+

PR #133, opened 2023-06-03
Dogfood finally passes :tada: 

xFrednet:
This PR fixes all the nits and adds all the expressions needed to make `cargo dogfood` pass. This almost finishes the work from #52, besides the `.await` expression.

There is not too much more to say. I thought about adding `cargo dogfood` to the CI, but I think it's better to wait with that, until we have the first custom lints. For now, this is a nice step into the right direction.

---

cc: https://github.com/rust-marker/marker/issues/52


xFrednet:
I want to rebase my current branch on these changes. As there hasn't been anyone interested in reviewing these changes, I think it's safe to just r+ it. This can always be reworked later.

bors r+

PR #136, opened 2023-06-04
Test: Use `emit_lint` instead of `println!` in tests

xFrednet:
Follow up of https://github.com/rust-marker/marker/pull/121

Not much more to say. I'll r+ this directly, as these are automatically generated and would be a pain to review.

r? @ghost


xFrednet:
Thank you Clippy, now again

bors r+

Issue #137, opened 2023-06-04
Rustc: Add semantic type backend for expressions


xFrednet:
A big concept while linting is *type checking* almost every lint, requires some context information including the type expressions. The next big step, for Marker, is a conversion backend from rustc's driver to the API type representation.

PR #138, opened 2023-06-11
API + Rustc: Start in semantic types

xFrednet:
This is the start of the semantic type representation for marker. The current implementation, covers the basic semantic types. Enough to generate a gigantic test file.

Reviews are as always very welcome! I'm happy to answer any questions as well :)

cc: https://github.com/rust-marker/marker/issues/137


xFrednet:
I'm merging this PR in favor of creating the next follow-up PR. The added code can always be updated later :)

bors r+

Issue #93, opened 2023-01-14
Get the `ItemId` based on a given ast-path


xFrednet:
During linting, a common pattern is to check for a specific type, method, or trait. In rustc these are identified using [`DefId`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/def_id/struct.DefId.html)/[`LocalDefId`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/def_id/struct.LocalDefId.html)'s or type relative checks for methods.

Marker currently doesn't provide a way to check for a specific item, as it only provides the `ItemId` and no further information. Ideally, the `AstContext` should provide a resolve method or something similar that takes a path like `std::vec::Vec` and returns the `ItemId` for comparison.

---

Clippy tries to avoid card coded paths in favor of diagnostic items. This makes sense in Clippy's context, especially when Clippy can just add the IDs to rustc types. See:
* https://github.com/rust-lang/rust-clippy/issues/5393
* https://rustc-dev-guide.rust-lang.org/diagnostics/diagnostic-items.html

For Marker, this is not a viable option, as the `rustc_diagnostic_item` attribute is compiler intern. Even if marker would add a similar attribute, users might not always be able to add these to items in question.

Hard-coded paths have one major problem: The user known path might only be a reexport of items from a different module. `std::iter::Iterator` for instance, is actually located at `std::iter::traits::iterator::Iterator`. If we require the user to specify the absolute path, the lints would break if the items would be moved internally. If we accept the public path, we have to resolve the item, which could potentially be expensive, especially for modules with glob imports.

Ideally, I'd like to choose the second option. Maybe we can try to reuse rustc's path resolution, to find the items in question?

xFrednet:
I'll start looking into this issue slowly. For reference, Clippy has [`clippy_utils::def_path_res`](https://doc.rust-lang.org/stable/nightly-rustc/clippy_utils/fn.def_path_res.html). The documentation notes, that a path like `std::vec` could resolve to the module and macro which is a good thing to keep in mind. I'm considering, if it might be good to restrict the resolve function to `TyDefId`s, as that should be the main usage of it anyways

xFrednet:
[...]

PR #140, opened 2023-06-16
API: Add interface to resolve `TyDefId` by a path string

xFrednet:
This is not the ideal implementation, as it requires manual path resolution. Rustc sadly hides the related functions from the public interface. https://github.com/rust-marker/marker/issues/93 provides additional information.

Closes: https://github.com/rust-marker/marker/issues/93


xFrednet:
I'm merging this to do some refactoring. We can always come back and update the code later :)

bors r+

PR #152, opened 2023-06-25
Maintanance: Update dependencies (where possible)

xFrednet:
Just some regular maintenance, nothing too interesting :)

There are still some older dependencies in the tree. I'm guessing from `cargo_fetch`? I've created https://github.com/Niki4tap/cargo_fetch/issues/1 to ask about the status of that project :)


xFrednet:
I've updated the description. Now next round

bors r+

PR #160, opened 2023-06-28
Set more `package` information in `Cargo.toml`s

xFrednet:
What the title says, nothing too interesting, but still important :). I'm looking forward to actually releasing these! :D


xFrednet:
Just a small rebase, I'll r+ this since there hasn't been any other activity

bors r+

Issue #163, opened 2023-07-07
Check "Rust API Guidelines" for Marker's API 


xFrednet:
[Rust API Guidelines](https://rust-lang.github.io/api-guidelines/about.html) contains a checklist with recommendations for designing a clean, idiomatic, and user-friendly API in Rust. Marker's API should be checked against these guidelines.

One important item are [sealed traits](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed) to prevent external crates from implementing Marker's API traits.

Issue #37, opened 2022-09-02
FFI related paper cuts in driver ↔ lint-crate communication


xFrednet:
When I created the prototype, I hacked together a basic *driver ↔ lint-crate* communication. After reworking the communication and learning more about FFI stuff, it turns out that the current implementation is unsound. This is a collection of all issues that need to be fixed:

* [ ] API types use non-`#[repr(C)]` types. These need to be replaced with FFI safe types or at least safely be transferred over the FFI boarder
    * Example: Slices, `Option<>`, `&str`
* [ ] Types from the FFI module should be hidden, as they are not part of the stable ABI
* [ ] Remove `&dyn Trait` pointers from the API
* [ ] Make sure that every exchanged type has `#[repr(C)]` (#6)

Note: We can change `#[repr(C)]` to a different ABI as long as it's stable like C

xFrednet:
I'm currently slowly working on these, while designing the AST. The last few PRs already made some good progress, like adding the FFI module to `linter_api`

xFrednet:
A lot of progress on this has been made in #54 as it removes a lot of `dyn` and traits. The API is not totally FFI safe, but has a good baseline by my estimation

PR #164, opened 2023-07-07
API: Follow "Rust API Guidelines", Seal traits :seal: and use enums instead of bools

xFrednet:
I generally checked the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/checklist.html). One of the most important changes is the new `Sealed` trait, to prevent lint crates from implementing Marker's API traits.

I've also added new `Mutability`, `Constness`, and `Syncness` enums, to remove boolean flags. I wanted to do this for some time, for the case that [keyword generics](https://blog.rust-lang.org/inside-rust/2022/07/27/keyword-generics.html) ever get added or some other magic. Having them shouldn't hurt, either way :)

Oh, and I also added a missing `#[repr(C)]`. That one should hopefully be the last missing one. (cc: https://github.com/rust-marker/marker/issues/37)

---

Closes: https://github.com/rust-marker/marker/issues/163


xFrednet:
I believe it's safe to merge this. :)

bors r+

Issue #167, opened 2023-07-10
Migrate away from `cargo_fetch` and bind to `cargo` directly


xFrednet:
`cargo-marker` currently uses [`cargo_fetch`](https://crates.io/crates/cargo_fetch) as a wrapper around [`cargo`](https://crates.io/crates/cargo). The dependency was added in #126 to support fetching crates from git and later from `crates.io`.

Sadly, `cargo_fetch` seems to be currently unmaintained: https://github.com/Niki4tap/cargo_fetch/issues/1

I think the best approach would be to remove that dependency and bind to `cargo` directly. That binding will require some maintenance with every release, but that's just a part of Marker, similar to updating the toolchain every release.

Hopefully, at some point we'll get: https://github.com/rust-lang/cargo/issues/1861

Issue #155, opened 2023-06-26
Separate `cargo_marker` CLI from the implementation 


xFrednet:
The CLI of `cargo_marker` is currently directly connected to the inner logic. This made the implementation easier. However, it would be nice to allow the usage of `cargo_marker` as a library as well. I'm mainly thinking of potential benefits, when it comes to uitest and the required setup.

This should also allow us to significantly speed up Marker's uitest. Currently, the lint crate is recompiled, every time.

Tasks:
* [x] Separate CLI from logic (Use structs instead if passing `clap::ArgMatches` around)
* [x] Expose important functions as a library
* [x] Migrate UI tests
* [ ] Document the magic

The library interface probably doesn't need to be stable. I'm happy to mentor this issue if someone is interested :)

Issue #60, opened 2022-11-24
Update cargo-linter and other CI to enable general use


xFrednet:
With most of the item backend implemented, it's finally time to run it on the first real crates and see what errors, todos and limitations come up. This is a collection of several paper cuts I ran into. Most of these are specific for the cargo interface.

* [ ] `cargo linter` passes `linter` as the second argument to `cargo-linter` which isn't handled correctly
* [ ] `cargo-linter` uses nightly features, which means that `cargo install ...` would require a nightly toolchain, this should be avoidable, at least for the cargo subcommand crate.
* [ ] `cargo-linter` starts the driver with the toolchain from the local `rust-toolchain` which can cause errors. Until we have the preimage to be a component in rustup, we need to manually handle the toolchain for the driver
* [ ] `cargo-linter` has a dev feature, which is turned on by default
* [ ] Lint crates are currently compiled with `cargo build` and not `cargo install` this requires nightly features and builds a debug binary
* [ ] The interface currently only supports local lint crates
* [ ] Lint crates should be specifiable in a file or something
* [ ] All of this needs to be documented.

PR #168, opened 2023-07-11
Cargo: Remove `cargo_fetch` dependency and rework the architecture

xFrednet:
This PR exploded a bit :sweat_smile:. Removing the dependency required a few more changes than expected. The good news are, that marker now uses a stable way to fetch lint crates which also supports registries directly. No more `cargo` dependency, just command calls. I'm happy :D

It also addresses or at least touches several other issues. Generally speaking, a nice update IMO. (The only big negative thing is the error handling. I've never learned how to do that properly, and cargo is getting more messy with less information after each refactoring :sweat_smile:. I'll create an issue for that :))

---

Closes: https://github.com/rust-marker/marker/issues/167
Closes: https://github.com/rust-marker/marker/issues/155
Closes: https://github.com/rust-marker/marker/issues/87

CC: https://github.com/rust-marker/marker/issues/60


xFrednet:
Btw, let's make sure it works on every system

bors try

xFrednet:
bors try

xFrednet:
bors try

xFrednet:
Windows...

bors try

xFrednet:
I noticed that the `cargo-marker` as a library for UI tests approach won't work for out of tree tests... Looks like I'll just have to revert to the old test setup for now :/

xFrednet:
And once again

bors try

xFrednet:
This was a fight... and still is... but for now:

bors try

xFrednet:
I think now it works... Damn that was a fight...

xFrednet:
I'm sure this is still broken somehow, but everything seems to be working. So let's follow the: "What's the worst that can happen ¯\_(ツ)_/¯" philosophy and merge this. I have another branch, which requires the changes from this one and we can always update the code later :)

bors r+

PR #173, opened 2023-07-14
Rustc: Warn when a user attempts to manually install the driver

xFrednet:
Just a small update, to prevent the accidental installation of rustc's driver :)


xFrednet:
Let's get this day started:

bors r+

PR #176, opened 2023-07-15
Rustc: Fix spans, sound and ICEs

xFrednet:
I ran Marker on several crates and fixed ICEs as they came up. I also found a soundness issue in the previous `SpanSource` implementation. Totally fixing spans is a task for another darker day, but this should work well enough for v0.1.0.

I tested Marker on the following crates:
* `clap`
* `parking_lot`
* `rand`
* `regex`
* `serde`
* `time`
* `tokio`
* `unicode-xid`

I want to test more crates, also as part of the CI, but this PR has grown too much already.


xFrednet:
Wait no rest for the wicket, bors if you may:

bors r+

Issue #10, opened 2022-04-29
Add panic hoock for


xFrednet:

xFrednet:
I'll work on the basic implementation of this :upside_down_face:

xFrednet:
This is actually a bit more complicated due to the `RefUnwindSafe` trait requirement of `std::panic::catch_unwind`. If we can't have a `catch_unwind` for every lint pass, I would at least like to add some additional information to the panic, like the current linted node etc. However, for now, I'll unassign myself and work on some other stuff

xFrednet:
From the [FFI](https://doc.rust-lang.org/nomicon/ffi.html) documentation, it looks like panics and unwinding is undefined across FFI, which this project currently uses. This means:

1. `catch_unwind` or a custom panic hook is required to ensure that all behavior is defined
2. The panic needs to be handled in the lint crate, this can't be done by the adapter.

In #30 I've reworked how the lint crate functions are called. That PR introduces macros to generate the entire boilerplate. `catch_unwind` could be implemented as part of the `export_lint_pass` macro

PR #180, opened 2023-07-16
Rustc: Add ICE hook and handle const generics

xFrednet:
The PRs are getting a bit out of hand, but at this point, I'm just grinding through any issues I know in the first come, first solved order. The goal is just to get ready for v0.1.0

CC: https://github.com/rust-marker/marker/issues/10


xFrednet:
I think this is a record for the shortest review time for a PR that should have been reviewed xD. Oh well, probably no time in waiting:

bors r+

PR #184, opened 2023-07-17
API: Add `Syn` prefix to syntactic generics and refactor the inner `generic` module

xFrednet:
This now aligns generics with the naming conventions of types. Marker could use a module called `syn` and `sem` instead. Personally, I like this version more, but this should probably be discussed before version v1.0.0

BTW: Just a big old refactoring, nothing really new was added in this one :)


xFrednet:
I added the `Syn` prefix to the syntactic parameters and squashed the commits. No reason to have 6x for just this refactoring. Now it should be ready:

bors r+

PR #191, opened 2023-07-18
Add several more checks to CI

Veetaha:
Some of this code was taken from [my public repo](https://github.com/Veetaha/snowpity/blob/master/.github/workflows/ci.yml), and some of the scripts were copied from my private repo (with the permission to do so gained as well).

## New checks to CI
- [`typos`](https://github.com/crate-ci/typos): this one has a static dictionary of known typos, and is famous for very low rate of false positives for that reason. It analyzes all text files in the repo and finds known typos in them. It's possible to autofix all typos with `typos -w`. 
- [`taplo`](https://github.com/tamasfe/taplo): formatter for TOML files just like `rustfmt`, it's possible to auto-format toml files with `taplo fmt` and configure the style as you wish in `taplo.toml`. 
- [`cargo-machete`](https://github.com/bnjbvr/cargo-machete) a heuristic check for unused dependencies. It parses the Rust files basically with a regex to figure out if the crate's name is ever mention in any paths and use statements in the sources. It is blazingly fast for that reason and nowhere near `cargo-udeps`, because it doesn't require compilation, `nightly` or anything like that. It's possible to ignore false-positive errors with the package metadata if you ever stumble into those.

## Scripts

All these checks are blazingly fast. I have them on my pre-commit hook in other repos, but I didn't add pre-commit hooks in this PR. 

I also added bash scripts for downloading the pre-compiled binaries for these dependencies, so it's possible to download them either locally or on CI with the same code. The downloading code is instrumented with good logging and exponential retries (in case github's CDN feels unwell) and good logging. This is how it looks on CI:

image

The sections marked with the 👉 symbol are expandable, they [use github's log groups ](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines)

If you ever want to download precompiled binaries for any other tools in the future it's easy to copy some existing download script in `scripts/download` for the tool you want to add and adapt the URL and `download_and_decompress` parameters for the delivery method of that tool.

## Refactorings

Replaced the `actions-rs` github actions for installing the toolchain with [Dtolnay's actions](https://github.com/dtolnay/rust-toolchain), because the former are deprecated and unmaintained, while Dtolnay's action is the modern maintained alternative.

Added `--no-deps` to `cargo doc`, this should speed up this CI step, because I am sure you don't want to build the docs for your dependencies on CI.

### DIFF
--- a/.github/workflows/rust_bors.yml
+++ b/.github/workflows/rust_bors.yml
@@ -44,9 +44,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: rm rust-toolchain
-    - uses: dtolnay/rust-toolchain@master
-      with:
-        toolchain: stable
+    - uses: dtolnay/rust-toolchain@stable
 
     - run: rustc -vV
 
--- a/.github/workflows/rust_bors.yml
+++ b/.github/workflows/rust_bors.yml
@@ -16,29 +16,25 @@ env:
 
 jobs:
   rust:
-    strategy:
-      matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
-
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
 
     # Setup
     steps:
     - uses: actions/checkout@v3
-    - uses: actions-rs/toolchain@v1
+    - uses: dtolnay/rust-toolchain@master
       with:
-          toolchain: nightly-2023-07-13
-          components: cargo, clippy, rustfmt
+        toolchain: nightly-2023-07-13
+        components: rustfmt, clippy
+
     - run: rustc -vV
-    - run: cargo build
-  
+
     # Tests
     - run: cargo test
     - run: cargo clippy
     - run: cargo fmt --check
-    - run: cargo doc
+    - run: cargo doc --no-deps
 
-  # This task ensures, required packages can be build with a stable toolchain
+  # This task ensures, required packages can be built with a stable toolchain
   # the only package requiring nightly should be `marker_rustc_driver` and
   # optionally `marker_adapter`
   rust-crates-build-stable:
@@ -48,10 +44,10 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: rm rust-toolchain
-    - uses: actions-rs/toolchain@v1
+    - uses: dtolnay/rust-toolchain@master
       with:
         toolchain: stable
-        components: cargo
+
     - run: rustc -vV
 
     # Test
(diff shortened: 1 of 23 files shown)


xFrednet:
Thank you for the PR and also requesting permission to use these scripts! I'll review it some time this week :)

scripts/download/cargo-machete.sh
```diff
@@ -0,0 +1,17 @@
...
+
+version=0.5.0
+
+base_url=https://github.com/bnjbvr/cargo-machete/releases/download/v$version
+file_stem=cargo-machete-v$version-$arch_rust-unknown-linux-musl
+
+download_and_decompress \
+    --check-hash sha256 \
+    $base_url/$file_stem.tar.gz \
+    --strip-components 1 $file_stem/cargo-machete
```
xFrednet, line 17: The [readme](https://github.com/bnjbvr/cargo-machete/) of `cargo-machete` suggests using

```
cargo install cargo-machete
``` 

Do you use this to directly download the binary, or is there another reason? Same question for the other download scripts :)

scripts/download/taplo.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+script_dir=$(readlink -f $(dirname $0))
+
+. $script_dir/lib.sh
+
+version=0.7.0
```
xFrednet, line 9: There appears to be a new version of this tool :)

```suggestion
version=0.8.1
```

xFrednet (commented):
This looks good to me, one question and one small update, then we can merge it. Thank you for the support :D

scripts/download/cargo-machete.sh
Veetaha: Downloading the pre-compiled binary is much faster than building it from source on every CI run. You may compare how much time it takes to `cargo install cargo-machete` with the time it takes to download. Hint, downloading it takes less then a second:
image

scripts/download/taplo.sh
Veetaha: Good catch. I somehow missed this upgrade, even though thought about it.

xFrednet (approved):
Looks good to me, thank you for the addition! I'm happy to have these new checks in Marker's CI :heart: 

bors r+

xFrednet:
bors retry

.github/workflows/rust_bors.yml
```diff
@@ -16,29 +16,25 @@ env:
 
 jobs:
   rust:
-    strategy:
-      matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
-
-    runs-on: ${{ matrix.os }}
```
xFrednet, line 23: The bors workflow uses a matrix, to test marker on ubuntu, windows and macos. Bors times out, as it was waiting on the `rust (macos-latest), ...` jobs. Could you readd the matrix? Then everything should work again :)

Veetaha: I see, I mistakenly removed it when copying. Will fix

xFrednet:
Thank you!

bors r+

Issue #181, opened 2023-07-16
API: The `Parameter` representation for callable objects should use a pattern


xFrednet:
The `Parameter` struct currently stores a name for every parameter instead of a pattern. However, Rust specifically allows patterns in parameters. This is simply an artifact from the time when patterns weren't implemented yet.

This requires adjustments in the API and Rustc driver 

* [x] FnItems: https://github.com/rust-marker/marker/pull/194
* [ ] Closure Expressions https://github.com/rust-marker/marker/pull/196

PR #196, opened 2023-07-24
API: Use patterns in `ClosureExpr` parameters

xFrednet:
This also removes the implementation of `CallableData` from `ClosureExpr`

Closes rust-marker/marker#181


xFrednet:
I still need to correct the return types of async functions, then this should be ready :)

marker_api/src/ast/expr.rs
```diff
@@ -377,7 +377,7 @@ mod test {
         assert_eq!(24, size_of::<CharLitExpr<'_>>(), "CharLitExpr<'_>");
         assert_eq!(24, size_of::<BoolLitExpr<'_>>(), "BoolLitExpr<'_>");
         assert_eq!(88, size_of::<BlockExpr<'_>>(), "BlockExpr<'_>");
-        assert_eq!(88, size_of::<ClosureExpr<'_>>(), "ClosureExpr<'_>");
+        assert_eq!(72, size_of::<ClosureExpr<'_>>(), "ClosureExpr<'_>");
```
Veetaha, line 380: I recommend you to use a snapshot-testing library.
I've been using [expect-test](https://docs.rs/expect-test/latest/expect_test/), which is a suprisingly low-overhead crate for snapshot testing.

This way you can capture all the data bout the types and their sizes in a string snapshot (inline our out-of-line in a separate file) and re-generate it if something changes with `UPDATE_EXPECT=1 cargo test`, which means maintaining tests becomes almost a zero-effort task, because they rewrite themselves, and you just review the diff.

This testing library was originally developed for use in `rust-analyzer` by [`@matklad`](https://github.com/matklad) and I've been using it in production successfuly so far. The one thing is that the library is barebones, and requires some small amount of batteries (like I added [in my repo here](https://github.com/Veetaha/snowpity/tree/master/crates/test-bat/src)) to make JSON snapshots and auto-format snapshots such that they don't exceed 80 lines in the editor

xFrednet:
> I still need to correct the return types of async functions, then this should be ready :)

I mixed up the PRs, this one is ready to be merged :)

marker_api/src/ast/expr.rs
xFrednet: Thank you for the suggestion. I created #202 to investigate this at one point :)

xFrednet:
I'll merge this now, to avoid too many conflict with coming PRs

bors r+

Issue #200, opened 2023-07-26
Lint attributes currently don't work for marker lints


xFrednet:
Extracted from #189:

> I tried everything I could think of, but `#[allow(...)]` doesn't work in any possible way. Here was my last non-working try
> 
> ```rust
> #![feature(register_tool)]
> #![register_tool(marker)]
> 
> fn main() {
>     let path = std::path::Path::new("foo.txt");
> 
>     #[allow(marker::to_string_on_cow_str)]
>     let _ = path.to_string_lossy().to_string();
> }
> ```
> [...]
> 
> The lint is not ignored, it's always reported when I run `cargo marker`

Marker version 0.1.1

PR #203, opened 2023-07-27
Rustc: Create `Adapter` early to register lints with rustc

xFrednet:
Registering the lints early, makes rustc track the lint level correctly (Assuming that marker is registered as a tool)

Not much more to say, I wrote this thing last night and I was super tired at the end. But it works, that's the important part ^^

---

Closes: rust-marker/marker#200

### DIFF
--- a/marker_adapter/src/lib.rs
+++ b/marker_adapter/src/lib.rs
@@ -7,10 +7,7 @@
 pub mod context;
 mod loader;
 
-use std::{
-    cell::{RefCell, RefMut},
-    ops::ControlFlow,
-};
+use std::{cell::RefCell, ops::ControlFlow};
 
 use loader::LintCrateRegistry;
 use marker_api::{
@@ -56,17 +53,13 @@ impl Adapter {
     }
 
     pub fn process_krate<'ast>(&self, cx: &'ast AstContext<'ast>, krate: &Crate<'ast>) {
-        // `RefMut::map` is used to get a *pure* mutable reference, which can
-        // then be used in the visitor. It might not be super pretty, but it works :)
-        RefMut::map(self.inner.borrow_mut(), |inner| {
-            inner.external_lint_crates.set_ast_context(cx);
+        let inner = &mut *self.inner.borrow_mut();
 
-            for item in krate.items() {
-                visitor::traverse_item::<()>(cx, inner, *item);
-            }
+        inner.external_lint_crates.set_ast_context(cx);
 
-            inner
-        });
+        for item in krate.items() {
+            visitor::traverse_item::<()>(cx, inner, *item);
+        }
     }
 }
 
--- a/marker_adapter/src/lib.rs
+++ b/marker_adapter/src/lib.rs
@@ -7,7 +7,10 @@
 pub mod context;
 mod loader;
 
-use std::ops::ControlFlow;
+use std::{
+    cell::{RefCell, RefMut},
+    ops::ControlFlow,
+};
 
 use loader::LintCrateRegistry;
 use marker_api::{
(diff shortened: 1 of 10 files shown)


marker_adapter/src/lib.rs
```diff
@@ -22,34 +25,52 @@ use marker_api::{
...
+        self.inner.borrow().external_lint_crates.collect_lint_pass_info()
     }
 
-    pub fn process_krate<'ast>(&mut self, cx: &'ast AstContext<'ast>, krate: &Crate<'ast>) {
-        self.external_lint_crates.set_ast_context(cx);
+    pub fn process_krate<'ast>(&self, cx: &'ast AstContext<'ast>, krate: &Crate<'ast>) {
+        // `RefMut::map` is used to get a *pure* mutable reference, which can
+        // then be used in the visitor. It might not be super pretty, but it works :)
+        RefMut::map(self.inner.borrow_mut(), |inner| {
+            inner.external_lint_crates.set_ast_context(cx);
```
Veetaha, line 62: This should work as well without `RefMut::map`
```rust
let inner = &mut *self.inner.borrow_mut();

inner.external_lint_crates.set_ast_context(cx);

for item in krate.items() {
    visitor::traverse_item::<()>(cx, inner, *item);
}
```

marker_rustc_driver/src/lint_pass.rs
```diff
@@ -1,25 +1,65 @@
...
 
+impl MarkerLintPass {
+    pub fn marker_lints() -> Vec<&'static Lint> {
+        ADAPTER.with(|adapter| {
+            let mut rustc_lints = vec![];
+            for info in adapter.lint_pass_infos() {
+                rustc_lints.extend_from_slice(info.lints());
+            }
+
+            rustc_lints
```
Veetaha, line 36:
```suggestion
             adapter
                .lint_pass_infos()
                .iter()
                .flat_map(marker_api::LintPassInfo::lints)
                .copied()
                .collect()
```

marker_rustc_driver/src/main.rs
```diff
@@ -49,12 +52,20 @@ struct MarkerCallback;
...
-        // Clippy explicitly calls any previous functions. This will not be
-        // done here to keep it simple and to ensure that only known code is
-        // executed.
+        // Clippy explicitly calls any previous `register_lints` functions. This
+        // will not be done here to keep it simple and to ensure that only known
+        // code is executed.
         assert!(config.register_lints.is_none());
 
         config.register_lints = Some(Box::new(|_sess, lint_store| {
+            // Register lints from lint crates. This is required to have rustc track
```
Veetaha, line 61: Are these `thread_local` required as a cache because the callbacks are invoked on per-crate basis, but the values they return for each crate are the same?

Maybe the adapter / converter / lints map, and other stuff could be a field on `MarkerLintPass`? This would avoid some of the thread locals

xFrednet: In Marker caches are often used to prevent translating existing notes again and leaking memory. This is a special case, though. Rustc assumes that all lints are static pointers. It uses the memory address as an ID and also for hashing etc. Marker is therefore to keep the memory addresses of these lints static.

The `MarkerLintPass` of the rustc driver is only conditionally constructed later. I tried around with storing it in there before, but then that instance also needed to be stored somewhere. This was the nicest solution I could come up with, but there are most likely better ones.

If you mean the `MarkerLintPass` of the adapter, that's a no, since the rustc lint instance is a concept of the rustc driver.

(The lint passes should probably have different names, now that I searched for it xD
)

NathanFrasier (approved):
While, I don't think I have quite the familiarity with the codebase to give a comprehensive review. I have given this a look and I don't see any problems.

xFrednet:
Thank you for the review, you two!

Reviewing PRs is also a great way to learn. 

bors r=NathanFrasier

PR #204, opened 2023-07-28
cargo: Pass the toolchain to all instances of MetadataCommand

### LINKED ISSUE #188: Error: NoTargetDir when running in a workspace with rust-toolchain `1.65.0` version
The error is simply
```
$ cargo marker
Error: NoTargetDir
```
Our cargo workspace is sitting on an older version of rust toolchain, although we are planning to upgrade soon, but the error is not informative, it doesn't show what actually went wrong. Only by inspecting the code of `cargo-marker` and experimenting I figured out the rust version was causing this.

To reproduce this you may create a crago workspace with `rust-toolchain` file with `1.65.0` in it.

The version of cargo-marker is `0.1.1` and `marker_lints` is `0.1.1` as well

NathanFrasier:
Closes #188

Tested using rust-toolchain 1.65.0

### DIFF
--- a/cargo-marker/src/backend/lints/fetch.rs
+++ b/cargo-marker/src/backend/lints/fetch.rs
@@ -17,7 +17,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use cargo_metadata::{Metadata, MetadataCommand};
+use cargo_metadata::Metadata;
 
 use crate::{backend::Config, config::LintDependencyEntry, ExitStatus};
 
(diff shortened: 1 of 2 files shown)


xFrednet (approved):
The changes look good to me, and everything works like a charm. Thank you very much for the update and nice solution!

bors r+

xFrednet:
@NathanFrasier Just a small FYI, the email you used in the commit, doesn't link to your GH profile. For Marker this is totally fine, so this is just a small note, in case this is not intentional. (Sorry, that I only noticed that after the merge :sweat_smile: )

NathanFrasier:
I'll fix it for future commits, but my emails have already been thoroughly linked across the Internet, not a huge deal to me.

Issue #209, opened 2023-08-01
Doc: Setup a mdbook to crate *The Marker Book*


xFrednet:
What the title says. I would like the book to be deployed using GitHub-Pages. I won't add too much detail to this issue, as I'll claim it directly.#

Side note: I contemplated calling it *The Textmarker*, but that is probably only funny to me and confusing for everyone else ^^

PR #210, opened 2023-08-02
Doc: Start *The Marker Book* and setup *GitHub-Pages*

xFrednet:
This was quite the journey. It turns out, that my HTML and CSS skills have not really improved, but this is good enough for a simple page that redirects people.

The book is currently just a skeleton. I plan to move some existing documentation into it in a later PR. This one is already touching too many different areas (CI, GH-Pages, Book).

Maybe some more context, why I added a separate index file, that just forwards the user. I assume that Marker will at one point allow us to generate documentation for lint crates. This will be a simple list of all lints contained in the crate, similar to [Clippy's lint list](https://rust-lang.github.io/rust-clippy/master/index.html). At that point, I would like to deploy the documentation for `marker_lints` on Marker's GH pages. Having an index file with links, allows us to have a single url (`rust-marker.github.io/marker/`) with subdirectories for the book and other documentation. For now, only `rust-marker.github.io/marker/book` will be populated.

I've also added some links to docs.rs to give the site a bit more purpose. And it's even responsive :D

I have it deployed on my repo, for testing, here is the link: https://xfrednet.github.io/marker/

---

Closes: #209


docs/index.html
```diff
@@ -0,0 +1,121 @@
+<!DOCTYPE html>
```
Veetaha, line 1: Beware of the hottest 🔥 frontend framework today https://htmx.org/

Veetaha (approved):

docs/index.html
xFrednet: That's cool and actually good to know. :)

xFrednet:
I'm already working on the follow-up PR. I'm merging that now, thank you for the review @Veetaha 

bors r=Veetaha

Issue #190, opened 2023-07-18
Avoid Cargo's caching if any lint crates changed


xFrednet:
Cargo includes an automatic caching mechanism, that avoids running the rustc diver if nothing changed. This mechanism doesn't account for potential changes inside the loaded lint crates.

AFAIK, it's possible to add more files to the cache check. If we add the compiled lint crates, it might just work itself out :)

This bug was extracted from: https://github.com/rust-marker/marker/issues/189

> ## Feedback loop
> 
> I had to run `cargo clean` each time I change the lint and want to check how it works. My feedback loop was:
> 
> 1. Change the custom lint's code.
> 2. Run `cargo clean`
> 3. Run `cargo marker`
> 
> 
> If I omit the step `2` I still see the output from the old version of my lint code.
> 
> I remember there was a different problem in `cargo-clippy` a long ago, which was also worked around with `cargo clean` to bust the incremental compilation cache ([rust-lang/rust-clippy#2604 (comment)](https://github.com/rust-lang/rust-clippy/issues/2604#issuecomment-423813352)). It isn't what happens here, but I hope `marker` won't have the same caveat.

Issue #98, opened 2023-01-19
Avoid cargo caching, if relevant files have been changed.


xFrednet:
Cargo implements some caching for rustc drivers by default. This can cause confusing, when cargo returns cached results instead of running the driver again. In Clippy, this happened for the `Cargo.toml` and `clippy.toml` files. Note that changes to `.rs` files will cause a rerun by default.

Ideally, marker should rerun, if relevant files have been changed. https://github.com/rust-lang/rust-clippy/pull/9707 might be a good example on how to implement this. The cache should also be avoided if the lint crates have changed, either via console arguments, or in the `Cargo.toml` file.

Additionally, it might be worth to also provide lint crates the ability to define relevant external files, that should cause a rerun. However, this could also be too driver-specific.

xFrednet:
See also #190

Issue #201, opened 2023-07-26
Add a way to set the lint level on the function/statement level, without the `register_tool` feature


xFrednet:
Rustc currently forbids lint names starting with a tool name, like `marker::example`. The `register_tool` feature (https://github.com/rust-lang/rust/issues/66079) can be used to add `marker` as a tool. However, this would require every create that wants to set lint levels on a function/statement level to use nightly.

Ideally, I would like the error, to be reduced to a normal lint trigger, until the `register_tool` feature is discussed, one distant day. However, Marker still needs a solution meanwhile for crates that can't update to a new version of rustc.

I think the following two solutions should be investigated:
1. Conditional compilation, similar or the same like: https://github.com/trailofbits/dylint#conditional-compilation
2. See how rustc handles a prefix, like `marker_` instead. This could be a viable solution, if this only triggers the `unknown_lint` lint of rustc

All working solutions should be documented :)

NathanFrasier:
I'll take a look at this, although my immediate feeling is I will end up in over my head. I'll document anything I find here, but I wouldn't be expecting a PR soon.

xFrednet:
Perfect! Next week, I plan to create some simpler issues, probably focussed on utilities for Marker, if that's something you would be interested in :). Any documentation for this one, would already be a big help :)

xFrednet:
For the second point in the issue description, note that that should only be an unstable opt-in option from the user side. Maybe via an environment value or setting in the configuration :)

Veetaha:
Not sure if you already knew it or if it helps, but beware that there is a hacky workaround that is definitely discouraged from anyone on the rustc developers team. But.. in the absence of a better solution it seems to be a lesser evil than installing a nightly toolchain.

`rustc` allows the usage of unstable features in the presence of [`RUSTC_BOOTSTRAP=1` environment variable](https://rustc-dev-guide.rust-lang.org/building/bootstrapping.html#complications-of-bootstrapping). This way you can activate unstable features using a stable toolchain.

NathanFrasier:
I really don't like that, because that would mean the feature would become available, but linted code would still need to use the unstable feature, so it's a really unpleasant user burden. I think that probably makes that approach a non-starter for this project, simply because the _user_ would also need to do it. we could maybe do some kind of macro to only use the feature on linting runs, but that just feels like the wrong approach.

[...]

PR #211, opened 2023-08-04
Rustc: Avoid caching by tracking lint crates and env vars

xFrednet:
This was a fight, it required some refactoring in `marker_adapter` to get the information, which lint crates should actually be tracked.

The process of invoking the compiler programmatically is also confusing, IMO. I don't know how anyone should be able to figure this out, without the existence of Clippy as a reference. Even with the reference and my tries to leave helpful comments, it took some time to get back into it.

If anyone is interested in reviewing this, I suggest just taking a look at the `register_tracked_files` function and the `marker_adapter` changes. The rest is first class spaghetti code which will hopefully work flawlessly:tm:

---

Testing this is also... a challenge, in the end I had to manually install the driver, open a different project and add a path lint crate. Then I changed the lint crate to confirm that it avoids the caching. Not really recommendable, oh well :)

---

Closes: https://github.com/rust-marker/marker/issues/98
Closes: https://github.com/rust-marker/marker/issues/190

### DIFF
--- a/marker_adapter/src/loader.rs
+++ b/marker_adapter/src/loader.rs
@@ -1,11 +1,10 @@
 use cfg_if::cfg_if;
 use libloading::Library;
-
 use marker_api::{interface::LintCrateBindings, AstContext};
-use marker_api::{LintPass, LintPassInfo};
-
+use marker_api::{LintPass, LintPassInfo, MARKER_API_VERSION};
 use std::ffi::{OsStr, OsString};
 use std::path::PathBuf;
+use thiserror::Error;
 
 use super::{AdapterError, LINT_CRATES_ENV};
 
@@ -87,22 +86,14 @@ pub struct LintCrateRegistry {
 }
 
 impl LintCrateRegistry {
-    /// # Panics
-    ///
-    /// Panics if a lint in the environment couldn't be loaded.
-    pub fn new(lint_crates: &[LintCrateInfo]) -> Self {
+    pub fn new(lint_crates: &[LintCrateInfo]) -> Result<Self, LoadingError> {
         let mut new_self = Self::default();
 
         for krate in lint_crates {
-            let lib = match LoadedLintCrate::try_from_info(krate.clone()) {
-                Ok(v) => v,
-                Err(err) => panic!("Unable to load `{}`, reason: {err:?}", krate.path.display()),
-            };
-
-            new_self.passes.push(lib);
+            new_self.passes.push(LoadedLintCrate::try_from_info(krate.clone())?);
         }
 
-        new_self
+        Ok(new_self)
     }
 
     pub(super) fn set_ast_context<'ast>(&self, cx: &'ast AstContext<'ast>) {
@@ -182,9 +173,7 @@ impl std::fmt::Debug for LoadedLintCrate {
 
 impl LoadedLintCrate {
     fn try_from_info(info: LintCrateInfo) -> Result<Self, LoadingError> {
-        let lib: &'static Library = Box::leak(Box::new(
-            unsafe { Library::new(&info.path) }.map_err(|_| LoadingError::FileNotFound)?,
-        ));
+        let lib: &'static Library = Box::leak(Box::new(unsafe { Library::new(&info.path) }?));
 
         let pass = LoadedLintCrate::try_from_lib(lib, info)?;
 
@@ -196,17 +185,20 @@ impl LoadedLintCrate {
         let get_api_version = {
             unsafe {
                 lib.get::<unsafe extern "C" fn() -> &'static str>(b"marker_api_version\0")
-                    .map_err(|_| LoadingError::MissingLintDeclaration)?
+                    .map_err(|_| LoadingError::MissingApiSymbol)?
             }
         };
-        if unsafe { get_api_version() } != marker_api::MARKER_API_VERSION {
-            return Err(LoadingError::IncompatibleVersion);
+        let krate_api_version = unsafe { get_api_version() };
+        if krate_api_version != MARKER_API_VERSION {
+            return Err(LoadingError::IncompatibleVersion {
+                krate_version: krate_api_version.to_string(),
+            });
         }
 
         // Load bindings
         let get_lint_crate_bindings = unsafe {
             lib.get::<extern "C" fn() -> LintCrateBindings>(b"marker_lint_crate_bindings\0")
-                .map_err(|_| LoadingError::MissingLintDeclaration)?
+                .map_err(|_| LoadingError::MissingBindingSymbol)?
         };
         let bindings = get_lint_crate_bindings();
 
@@ -218,9 +210,14 @@ impl LoadedLintCrate {
     }
 }
 
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum LoadingError {
-    FileNotFound,
-    IncompatibleVersion,
-    MissingLintDeclaration,
+    #[error("the lint crate could not be loaded: {0:#?}")]
+    LibLoading(#[from] libloading::Error),
+    #[error("the loaded crated doesn't contain the `marker_api_version` symbol")]
+    MissingApiSymbol,
+    #[error("the loaded crated doesn't contain the `marker_lint_crate_bindings` symbol")]
+    MissingBindingSymbol,
+    #[error("incompatible api version:\n- lint-crate api: {krate_version}\n- driver api: {MARKER_API_VERSION}")]
+    IncompatibleVersion { krate_version: String },
 }
--- a/marker_adapter/src/loader.rs
+++ b/marker_adapter/src/loader.rs
@@ -5,6 +5,9 @@ use marker_api::{interface::LintCrateBindings, AstContext};
 use marker_api::{LintPass, LintPassInfo};
 
 use std::ffi::{OsStr, OsString};
+use std::path::PathBuf;
+
+use super::{AdapterError, LINT_CRATES_ENV};
 
 /// Splits [`OsStr`] by an ascii character
 fn split_os_str(s: &OsStr, c: u8) -> Vec<OsString> {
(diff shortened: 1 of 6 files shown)


marker_rustc_driver/src/main.rs
```diff
@@ -71,6 +95,49 @@ impl rustc_driver::Callbacks for MarkerCallback {
...
+    // At this point, the lint crate paths should be absolute. Therefore, we
+    // can just throw it in the `files` set.
+    for lint_crate in lint_crates {
+        if let Some(name) = lint_crate.path.to_str() {
+            files.insert(Symbol::intern(name));
+        } else {
+            // The name is not a valid UTF-8 String. This is not ideal but at the
+            // same time not problematic enough to throw an error. It just means
+            // that the lint crate can't be tracked and the driver might reuse
+            // a cache if only that crate was changed.
```
Veetaha, line 128: I recommend you not to care about supportin non-utf8 paths. There are a bunch of reasons to do so. First of all, it simplifies the code, and second a lot of tools assume they work with UTF8 paths, for example cargo-metadata crate does so. You may see that it uses [`Utf8Path[Buf]`](https://docs.rs/cargo_metadata/latest/cargo_metadata/struct.Metadata.html#structfield.target_directory) for storing all the paths parsed from `cargo metadata`.

My first impression when I saw this in `cargo-metadata` was "why do you limit the users to UTF8", but then I read the reasoning from [`camino`](https://github.com/camino-rs/camino) crate that implements the `Utf8Path[Buf]`, and was convinced that this is the right thing to do. I never in my life had to work with non-UTF8 paths. If you ever happen to work with something like that you are already fucked, and cargo-marker won't be the only tool that doesn't work for you. Can you imagine someone invoking `cargo-marker` on a project that uses invalid UTF8 in its code structure? I can't, and thus I propose adopting `camino` and releaving this mental complexity from the codebase. My question is wether you would accept a PR with such kind of change?

marker_rustc_driver/src/main.rs
```diff
@@ -71,6 +95,49 @@ impl rustc_driver::Callbacks for MarkerCallback {
...
+    let cargo_file = Path::new("./Cargo.toml");
+    if cargo_file.exists() {
+        files.insert(Symbol::intern("./Cargo.toml"));
+    }
+
+    // At this point, the lint crate paths should be absolute. Therefore, we
+    // can just throw it in the `files` set.
+    for lint_crate in lint_crates {
+        if let Some(name) = lint_crate.path.to_str() {
+            files.insert(Symbol::intern(name));
```
Veetaha, line 123: I wonder if it track the directory recursively. IIRC there was some ~~bug~~ feature in cargo that if your use `rerun-if-changed` on a directory in a build script, then only the dir metadata or just the first level of entries will be tracked.

marker_rustc_driver/src/main.rs
```diff
@@ -71,6 +95,49 @@ impl rustc_driver::Callbacks for MarkerCallback {
...
+            // The name is not a valid UTF-8 String. This is not ideal but at the
+            // same time not problematic enough to throw an error. It just means
+            // that the lint crate can't be tracked and the driver might reuse
+            // a cache if only that crate was changed.
+        }
+    }
+
+    // Track the driver executable in debug builds
+    #[cfg(debug_assertions)]
+    if let Ok(current_exe) = env::current_exe()
```
Veetaha, line 134: `if let Ok(_)` patterns always catch my eye, because they silently ignore an error. I'd rather log the error if it happens. But, for that there should be `tracing` installed

marker_rustc_driver/src/main.rs
```diff
@@ -121,22 +188,20 @@ fn toolchain_path(home: Option<String>, toolchain: Option<String>) -> Option<Pat
...
-
-Other options are the same as `cargo check`.
+        --rustc              Pass all arguments to rustc
+    -V, --version            Print version information and exit
+        --toolchain          Print the required toolchain and API version
 
 ---
 
-This message belongs to a specific driver, if possible you should avoid
+This message belongs to a specific  marker driver, if possible you should avoid
```
Veetaha, line 204:
```suggestion
This message belongs to a specific marker driver, if possible you should avoid
```

marker_rustc_driver/src/main.rs
```diff
@@ -170,14 +240,14 @@ fn main() {
             orig_args.extend(vec!["--sysroot".into(), sys_root]);
         };
 
-        // make "marker-driver --rustc" work like a subcommand that passes further args to "rustc"
-        // for example `marker-driver --rustc --version` will print the rustc version that marker-driver
-        // uses
+        // make "marker_rustc_driver --rustc" work like a subcommand that passes
```
Veetaha, line 243: I see that previously dashes were used to separate package names. I don't understand why the choice was made to use underscores. To my opinion, crate and binary names should have kebab case. This is a very common convention. For example, cargo plugins are enforced to use kebab-case, tools like `rust-gdb`, `rust-gdbui`, `rust-lldb`, `typos-cli`. It's also confusing that you have to install `cargo_marker`, but then it is insalled as `cargo-marker`.

marker_rustc_driver/src/lint_pass.rs
```diff
@@ -18,17 +18,26 @@ thread_local! {
...
 
 pub struct RustcLintPass;
 
 impl RustcLintPass {
+    pub fn init_adapter(lint_crates: &[LintCrateInfo]) -> Result<(), AdapterError> {
+        ADAPTER.with(move |cell| {
+            if cell.get().is_none() {
+                cell.set(Adapter::new(lint_crates)?).unwrap();
+            };
+            Ok(())
```
Veetaha, line 32:
```suggestion
            cell.get_or_try_init(|| Adapter::new(lint_crates))?;
            Ok(())
```

Veetaha (approved):
I didn't notice the "spaghetti" that you was referring to

marker_rustc_driver/src/main.rs
NathanFrasier: I second the notion here, if cargo metadata already assumes utf-8 paths, we should lean on that for simplicity. I think this is the right design decision.

marker_rustc_driver/src/main.rs
NathanFrasier: Agreed, I can't tell you how many times I've typed the kebab case version when I need the underscore version, this seems relatively simple to change if we want to. @xFrednet might have a good reason for changing it though.

xFrednet: :sweat_smile: Yeah, so, the background of this is that clippy uses underscores in their crates (`clippy_lints` & `clippy_utils`) for library crates. For library crates, this also makes sense IMO, since the `use` statement also uses underscores. I also looked at some other crates and there seemed to be no real consensus, which should be used bevy for example also always uses underscores.

Now for the binaries... when I decided on the name, I just reserved the `driver` and `cargo_marker` name without contemplating that these are binaries. Having underscores everywhere just seemed consistent. That was until I found out that Cargo expects the binary to use a dash instead of an underscore.

So, this basically comes down to, little experience when it comes to actually releasing crates/binaries and only learning about my mistakes after it was too late. I even emailed crates.io about maybe changing the name, but they said that this is not possible.

We could rename the binary in the same way that the `cargo_marker` crate is renamed, but the names are sadly stuck now. (The only other option, I see, would be to rename the project, which I don't want to)

marker_rustc_driver/src/main.rs
xFrednet: I'm fine with moving to UTF-8 based paths. If we document it, it should be a fine restriction to make. Though I would not do this in this PR

marker_adapter/src/loader.rs
```diff
@@ -40,53 +42,58 @@ fn windows_split_os_str(s: &OsStr, c: u8) -> Vec<OsString> {
...
+            }
 
-        // FIXME: Create issue for lifetimes and fix dropping and pointer decl stuff
+            list.push(LintCrateInfo {
+                path: PathBuf::from(entry_str),
+            });
+        }
 
-        Ok(pass)
+        Ok(list)
```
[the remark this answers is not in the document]
xFrednet: Yeah, these for loops often just arise, when I draft code and refactor it over and over. :+1:

xFrednet:
> I didn't notice the "spaghetti" that you was referring to

I feel like the control flow of the main function/module and the data flow in the function is a bit convoluted. There are several conditions, that can cause the flow to take another route or just terminate early. I'm just to very linear and deterministic code, so this might feel more normal to you :)

marker_adapter/src/loader.rs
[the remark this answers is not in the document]
xFrednet: > Also maybe this should use the [doc.rust-lang.org/stable/std/env/fn.split_paths.html](https://doc.rust-lang.org/stable/std/env/fn.split_paths.html) format?

Yep, just never knew about that one. I expect that this format will change soon, to also include the lint crate name for #201, but for now we can use that function :+1: And then it would probably be smart to use Serde to serialize and deserialize these entries.

cc: @NathanFrasier since this references #201

marker_rustc_driver/src/main.rs
xFrednet: I've recently talked to someone from a project that always uses the rustc/clippy driver directly. For Marker they would do the same. In such a case, a lint crate might be specified that is not inside the lints directory of the project.

Doing it this way doesn't put any restrictions on the user for little to no cost to us :)

marker_rustc_driver/src/main.rs
Veetaha: I see, that's unfortunate, but I don't see a better solution than renaming

marker_rustc_driver/src/main.rs
xFrednet: I can use unwrap, instead, either way, I think this is fine, since this is only enabled for debug builds.

xFrednet: I'll change it to a match

marker_rustc_driver/src/main.rs
xFrednet: There is one solution for the binaries: Make Marker successful enough that it's actually distributed with rustup instead ^^

xFrednet:
Thank you for the review and comments! It's really a big help and I appreciate it!

xFrednet:
I think it's safe to r+ this

bors r=Veetaha

Issue #175, opened 2023-07-14
Rework spans source information and add more utility functions for spans


xFrednet:
Span are a fundamental part of a linting interface. They define what should be shown in diagnostic messages and where to apply suggestions. Marker currently only has a limited `Span` implementation and a simple `snippet` function. Clippy and rustc both have a collection of util function to make it easier to deal with spans. It might be cool for Marker to have something similar.

See:
* rustc's [`SourceMap`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_span/source_map/struct.SourceMap.html#)
* Clippy's [`clippy_utils::source`](https://doc.rust-lang.org/stable/nightly-rustc/clippy_utils/source/index.html)

This util might need to be part of the `marker_api` crate to allow driver callbacks, similar to how the `AstContext` currently works :)

xFrednet:
Spans in general should be updated to include more information about where they came from, like from which macro and if they have parent spans. See the fixme comment above the `SpanSource` struct.

xFrednet:
I'll work on the source information and, in general, making spans more rounded. :)

xFrednet:
Just understanding all the variations of spans and macro expansion and documenting my current understanding is taking a significant amount of time. I never knew that spans were this complicated. In Clippy, we generally just abort if the code originates from a macro.

Oh well, when I have a working branch, this will probably be the best documented part of the API. Right now, it has about 150 lines of new doc comments. ^^

xFrednet:
I've reworked the spans in #218, Marker should still include more util methods, but for now, I consider this issue to be resolved :)

I'll try create issues for more util methods soon :)

Issue #189, opened 2023-07-18
My feedback for 0.1.1


Veetaha:
First of all, that's a great start! Writing custom lints feels rather nice for me, and doesn't need to depend on nightly at least from the interface side.

I used the `0.1.1` tooling to implement a lint that suggests using `into_owned()` instead of `to_string()` for `Cow<'_, str>`.

Here was my code:

<details>
<summary>Details</summary>

```rust
use marker_api::ast::generic::SemGenericArgKind;
use marker_api::ast::ty::SemTyKind;
use marker_api::prelude::*;
use marker_api::{LintPass, LintPassInfo, LintPassInfoBuilder};

#[derive(Default)]
struct ToStringOnCowStrPass {}
marker_api::export_lint_pass!(ToStringOnCowStrPass);

marker_api::declare_lint! {
    /// # What it does
    /// This lint detects the usage of `.to_string()` on a `Cow<str>`.
    ///
    /// # Example
    /// ```rs
    /// PathBuf::from("foo").to_string_lossy().to_string()
    /// ```
    ///
    /// Use instead:
    /// ```rs
    /// PathBuf::from("foo").to_string_lossy().into_owned()
    /// ```
    ///
    /// If you intended to clone the `Cow`, then use `.clone()` or `.to_owned()`.
    TO_STRING_ON_COW_STR,
    Warn,
}

impl LintPass for ToStringOnCowStrPass {
    fn info(&self) -> LintPassInfo {
        LintPassInfoBuilder::new(Box::new([TO_STRING_ON_COW_STR])).build()
    }

    fn check_expr<'ast>(
        &mut self,
        ctx: &'ast AstContext<'ast>,
        expr: marker_api::ast::expr::ExprKind<'ast>,
    ) {
        let ExprKind::Method(method) = expr else {
            return;
        };

        if method.method().ident().name() != "to_string" {
            return;
        }

        let SemTyKind::Adt(reciever) = method.receiver().ty() else {
            return;
        };

        if !ctx
            .resolve_ty_ids("std::borrow::Cow")
            .contains(&reciever.def_id())
        {
            return;
        }

        let [SemGenericArgKind::Ty(str)] = reciever.generics().args() else {
            return;
        };

        if !matches!(str, SemTyKind::Text(str) if str.is_str()) {
            return;
        }

        ctx.emit_lint(
            TO_STRING_ON_COW_STR,
            method.receiver().id(),
            "replace this with .into_owned()",
            method.method().ident().span(),
[...]

Issue #32, opened 2022-08-08
General discussion about rust-analyzer as a driver


HKalbasi:
For using another driver instead of rustc (like rust-analyzer) it should not depend on rustc representation of hir and ast, and ideally it should work on stable toolchain, without `#![feature(rustc_private)]`. Is there a plan for doing so? And how we can help in that direction?

xFrednet:
This is definitely planned. The `liner_api` crate which will contain the AST (like) representation of the code is based on the [Rust Reference](https://doc.rust-lang.org/stable/reference/). Only `linter_driver_rustc` lints with rustc internals and translates them to `liner_api` types. Which will then be passed over to the specific lint crates.

I know that rust-analyzer had a similar plan. However, that has stalled a bit AFAIK. I wanted to ask them on the status, but haven't gotten to that yet. :sweat_smile: 

> And how we can help in that direction?

This project is currently a bit chaotic, since the infrastructure etc hasn't been fully designed yet. So there are a lot of loose ends which are not yet documented well. I plan to write a small contribution guide and issues once the base infrastructure is done (Planned for the end of the month). Until then, you could help with code reviews (Currently, I make one larger PR per week or so). Or try to do some coding, but that might cause a lot of merge conflict until the basic structure is in place.

If you don't mind me asking, what is your background? Have you worked on rust-analyzer? And what do you enjoy doing in Rust? :upside_down_face:

HKalbasi:
Ah, so I was completely wrong, and the parts that are actually needed for implementing another driver (`linting_adapter` and `linting_api`) work in stable. I did some experiments and it looks very promising.

> I know that rust-analyzer had a similar plan. However, that has stalled a bit AFAIK. I wanted to ask them on the status, but haven't gotten to that yet.

What plan do you mean? I only know about shared parser library, which would make the `ast` equal between rustc and r-a, but it is no where near complete AFAIK (and it wouldn't help in HIR part, at least immediately).

> If you don't mind me asking, what is your background? Have you worked on rust-analyzer? And what do you enjoy doing in Rust?

I made some contributions in r-a, mostly around const generic / const eval and LSIF (this is a format for dumping intellisense data about a code repository) and I'm mostly familiar with r-a internals. The duplicate efforts between r-a and rustc makes me sad, so I'm really interested in librarification projects.

[...]

PR #218, opened 2023-08-18
API, Rustc: Rework spans (and fix a few errors)

xFrednet:
Spans are a fundamental part of linting in Rust. The previous implementation of spans in Marker were hacked together, with me always adding something new when it was needed. This PR reworks spans from the ground up.

The implementation is inspired by rustc's [`Span`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_span/struct.Span.html) and connected data structures. 

While testing, I also found a few ICEs and mistakes, that were fixed as well. The short summary:
* Marker's `Span`s are now cool
* Marker has less errors and panics
* Changes are almost net negative

An awesome effect of this change is that the `Debug` format implementation of spans looks a lot better now, like this:
```
path/file.rs:17:6 - 17:7,
```

---

cc: rust-marker/marker#175

### DIFF
--- a/marker_adapter/src/context.rs
+++ b/marker_adapter/src/context.rs
@@ -95,6 +95,8 @@ extern "C" fn span_snippet<'ast>(data: &'ast (), span: &Span<'ast>) -> ffi::FfiO
     unsafe { as_driver_cx(data) }.span_snippet(span).map(Into::into).into()
 }
 
+// False positive because `SpanSource` is non-exhaustive
+#[allow(improper_ctypes_definitions)]
 extern "C" fn span_source<'ast>(data: &'ast (), span: &Span<'_>) -> SpanSource<'ast> {
     unsafe { as_driver_cx(data) }.span_source(span)
 }
--- a/marker_adapter/src/context.rs
+++ b/marker_adapter/src/context.rs
@@ -5,7 +5,8 @@ use marker_api::{
     ast::{
         item::{Body, ItemKind},
         ty::SemTyKind,
-        BodyId, ExprId, ItemId, Span, SpanId, SymbolId, TyDefId,
+        BodyId, ExpnId, ExpnInfo, ExprId, FileInfo, FilePos, ItemId, Span, SpanId, SpanPos, SpanSource, SymbolId,
+        TyDefId,
     },
     context::DriverCallbacks,
     diagnostic::{Diagnostic, EmissionNode},
@@ -47,6 +48,9 @@ impl<'ast> DriverContextWrapper<'ast> {
             expr_ty,
             span,
             span_snippet,
+            span_source,
+            span_pos_to_file_loc,
+            span_expn_info,
             symbol_str,
             resolve_method_target,
         }
@@ -55,7 +59,7 @@ impl<'ast> DriverContextWrapper<'ast> {
 
 // False positive because `EmissionNode` are non-exhaustive
 #[allow(improper_ctypes_definitions)]
-extern "C" fn lint_level_at(data: &(), lint: &'static Lint, node: EmissionNode) -> Level {
+extern "C" fn lint_level_at<'ast>(data: &'ast (), lint: &'static Lint, node: EmissionNode) -> Level {
     unsafe { as_driver_cx(data) }.lint_level_at(lint, node)
 }
 
@@ -91,11 +95,27 @@ extern "C" fn span_snippet<'ast>(data: &'ast (), span: &Span<'ast>) -> ffi::FfiO
     unsafe { as_driver_cx(data) }.span_snippet(span).map(Into::into).into()
 }
 
+extern "C" fn span_source<'ast>(data: &'ast (), span: &Span<'_>) -> SpanSource<'ast> {
+    unsafe { as_driver_cx(data) }.span_source(span)
+}
+
+extern "C" fn span_pos_to_file_loc<'ast>(
+    data: &'ast (),
+    file: &FileInfo<'ast>,
+    pos: SpanPos,
+) -> ffi::FfiOption<FilePos<'ast>> {
+    unsafe { as_driver_cx(data) }.span_pos_to_file_loc(file, pos).into()
+}
+
+extern "C" fn span_expn_info<'ast>(data: &'ast (), expn_id: ExpnId) -> FfiOption<&'ast ExpnInfo<'ast>> {
+    unsafe { as_driver_cx(data) }.span_expn_info(expn_id).into()
+}
+
 extern "C" fn symbol_str<'ast>(data: &'ast (), sym: SymbolId) -> ffi::FfiStr<'ast> {
     unsafe { as_driver_cx(data) }.symbol_str(sym).into()
 }
 
-extern "C" fn resolve_method_target(data: &(), id: ExprId) -> ItemId {
+extern "C" fn resolve_method_target<'ast>(data: &'ast (), id: ExprId) -> ItemId {
     unsafe { as_driver_cx(data) }.resolve_method_target(id)
 }
 
@@ -117,7 +137,10 @@ pub trait DriverContext<'ast> {
 
     fn expr_ty(&'ast self, expr: ExprId) -> SemTyKind<'ast>;
     fn span(&'ast self, owner: SpanId) -> &'ast Span<'ast>;
-    fn span_snippet(&'ast self, span: &Span<'ast>) -> Option<&'ast str>;
+    fn span_snippet(&'ast self, span: &Span<'_>) -> Option<&'ast str>;
+    fn span_source(&'ast self, span: &Span<'_>) -> SpanSource<'ast>;
+    fn span_expn_info(&'ast self, expn_id: ExpnId) -> Option<&'ast ExpnInfo<'ast>>;
+    fn span_pos_to_file_loc(&'ast self, file: &FileInfo<'ast>, pos: SpanPos) -> Option<FilePos<'ast>>;
     fn symbol_str(&'ast self, api_id: SymbolId) -> &'ast str;
     fn resolve_method_target(&'ast self, id: ExprId) -> ItemId;
 }
--- a/marker_api/src/ast/common/id.rs
+++ b/marker_api/src/ast/common/id.rs
@@ -70,7 +70,7 @@ new_id! {
 }
 
 new_id! {
-    /// This ID uniquely identifies a user defined type during linting.
+    /// This ID uniquely identifies a macro during linting.
     pub MacroId: u64
 }
 
--- a/marker_api/src/ast/common/id.rs
+++ b/marker_api/src/ast/common/id.rs
@@ -125,6 +125,15 @@ new_id! {
 new_id! {
     /// **Unstable**
     ///
+    /// This id is used to identify a specific expansion. This type is only intended for internal
+    /// use. For now it's only intended for drivers to map spans back
+    #[cfg_attr(feature = "driver-api", visibility::make(pub))]
+    pub(crate) ExpnId: u64
+}
+
+new_id! {
+    /// **Unstable**
+    ///
     /// This id is used to identify symbols. This type is only intended for internal
     /// use. Lint crates should always get [`String`] or `&str`.
     #[cfg_attr(feature = "driver-api", visibility::make(pub))]
--- a/marker_api/src/ast/common/id.rs
+++ b/marker_api/src/ast/common/id.rs
@@ -70,6 +70,11 @@ new_id! {
 }
 
 new_id! {
+    /// This ID uniquely identifies a user defined type during linting.
+    pub MacroId: u64
+}
+
+new_id! {
     /// This ID uniquely identifies a body during linting.
     pub BodyId: u64
 }
--- a/marker_api/src/ast/common/span.rs
+++ b/marker_api/src/ast/common/span.rs
@@ -2,7 +2,7 @@ use std::marker::PhantomData;
 
 use crate::{context::with_cx, diagnostic::Applicability, ffi};
 
-use super::{MacroId, SpanId, SpanSrcId, SymbolId};
+use super::{ExpnId, MacroId, SpanId, SpanSrcId, SymbolId};
 
 /// A byte position used for the start and end position of [`Span`]s.
 ///
@@ -136,7 +136,7 @@ impl SpanPos {
 #[derive(Debug)]
 pub struct ExpnInfo<'ast> {
     _lifetime: PhantomData<&'ast ()>,
-    parent: SpanSrcId,
+    parent: ExpnId,
     call_site: SpanId,
     macro_id: MacroId,
 }
@@ -144,8 +144,7 @@ pub struct ExpnInfo<'ast> {
 impl<'ast> ExpnInfo<'ast> {
     #[must_use]
     pub fn parent(&self) -> Option<&ExpnInfo<'ast>> {
-        // TODO(xFrednet): Request info from Driver
-        todo!()
+        with_cx(self, |cx| cx.span_expn_info(self.parent))
     }
 
     /// The [`Span`] that invoked the macro, that this expansion belongs to.
@@ -162,7 +161,7 @@ impl<'ast> ExpnInfo<'ast> {
 #[cfg(feature = "driver-api")]
 impl<'ast> ExpnInfo<'ast> {
     #[must_use]
-    pub fn new(parent: SpanSrcId, call_site: SpanId, macro_id: MacroId) -> Self {
+    pub fn new(parent: ExpnId, call_site: SpanId, macro_id: MacroId) -> Self {
         Self {
             _lifetime: PhantomData,
             parent,
@@ -170,28 +169,6 @@ impl<'ast> ExpnInfo<'ast> {
             macro_id,
         }
     }
-}
-
-// FIXME(xFrednet): This enum is "limited" to say it lightly, it should contain
-// the more information about macros and their expansion etc. This covers the
-// basic use case of checking if a span comes from a macro or a file. The rest
-// will come in due time. Luckily it's not a public enum right now.
-//
-// See: rust-marker/marker#175
-#[repr(C)]
-#[allow(clippy::exhaustive_enums)]
-#[derive(Debug, Eq, PartialEq, Hash)]
-#[cfg_attr(feature = "driver-api", visibility::make(pub))]
-enum SpanSource<'ast> {
-    /// The span comes from a file
-    File(ffi::FfiStr<'ast>),
-    /// The span comes from a macro.
-    Macro(SpanSrcId),
-    /// The span belongs to a file, but is the result of desugaring, they should
-    /// be handled like normal files. This is variant mostly important for the driver.
-    // FIXME(xFrednet): All desugars are usually resugared by Marker, the driver should
-    // pass the span of the original sugared expression to the API.
-    Sugar(ffi::FfiStr<'ast>, SpanSrcId),
 }
 
 /// A region of code, used for snipping, lint emission, and the retrieval of
@@ -224,7 +201,7 @@ pub struct Span<'ast> {
     /// The source of this [`Span`]. The id space and position distribution is
     /// decided by the driver. To get the full source information it might be
     /// necessary to also pass the start and end position to the driver.
-    source: SpanSrcId,
+    source_id: SpanSrcId,
     /// This information could also be retrieved, by requesting the [`ExpnInfo`]
     /// of this span. However, from looking at Clippy and rustc lints, it looks
     /// like the main interested is, if this comes from a macro expansion, not from
@@ -361,15 +338,20 @@ impl<'ast> Span<'ast> {
         new_span.set_end(end);
         new_span
     }
+
+    #[must_use]
+    pub fn source(&self) -> SpanSource<'ast> {
+        with_cx(self, |cx| cx.span_source(self))
+    }
 }
 
 #[cfg(feature = "driver-api")]
 impl<'ast> Span<'ast> {
     #[must_use]
-    pub fn new(source: SpanSrcId, from_expansion: bool, start: SpanPos, end: SpanPos) -> Self {
+    pub fn new(source_id: SpanSrcId, from_expansion: bool, start: SpanPos, end: SpanPos) -> Self {
         Self {
             _lifetime: PhantomData,
-            source,
+            source_id,
             from_expansion,
             start,
             end,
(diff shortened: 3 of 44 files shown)


marker_rustc_driver/src/conversion/marker/common/span.rs
```diff
@@ -0,0 +1,70 @@
...
+        )
+    }
+
+    pub fn to_span_pos(&self, byte_pos: rustc_span::BytePos) -> SpanPos {
+        SpanPos::new(byte_pos.0)
+    }
+
+    pub fn to_span_source(&self, rust_span: rustc_span::Span) -> SpanSource<'ast> {
+        let ctxt = rust_span.ctxt();
+        if ctxt.is_root() {
```
Veetaha, line 26: With an early return the code should be one indentation level closer to the left side of the screen.

```rust
if !ctxt.is_root() {
    return SpanSource::Macro(self.alloc(self.to_expn_info(&ctxt.outer_expn_data())));
}

// The rest of the code here but without the indentation

```

marker_rustc_driver/src/conversion/marker/common/span.rs
```diff
@@ -0,0 +1,70 @@
...
+            let src_file = self.rustc_cx.sess.source_map().lookup_source_file(rust_span.lo());
+            let name = match &src_file.name {
+                rustc_span::FileName::Real(
+                    rustc_span::RealFileName::LocalPath(file_path)
+                    | rustc_span::RealFileName::Remapped {
+                        virtual_name: file_path,
+                        ..
+                    },
+                ) => file_path.to_string_lossy().to_string(),
```
Veetaha, line 35: This was acutally the lint I wrote as part of https://github.com/rust-marker/marker/issues/189 (`to_string_on_cow_str`).

You should never write `to_string_lossy().to_string()` write this instead, because it consumes the `Cow` and doesn't allocate when it already owned:

```suggestion
                ) => file_path.to_string_lossy().into_owned(),
```

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
+    /// ```
+    ///
+    /// There are several reasons, why a snippet might be unavailable. Also
+    /// depend on the used driver. You can also checkout the other snippet
+    /// methods to better deal with these cases:
+    /// * [`snippet_or`](Self::snippet_or)
+    /// * [`snippet_with_applicability`](Self::snippet_with_applicability)
+    #[must_use]
+    pub fn snippet(&self) -> Option<&'ast str> {
+        with_cx(self, |cx| cx.span_snipped(self))
```
Veetaha, line 262: `[off-topic]`
By reading the `rustc` source code I happened to find the crate [`scoped-tls`](https://docs.rs/scoped-tls/latest/scoped_tls/) created by rustc devs. 

Maybe it could be used as backend for storing the AST context? It would remove some `unsafe` code from the guts of `with_cx`

marker_rustc_driver/src/conversion/marker/common.rs
```diff
@@ -70,6 +73,15 @@ impl<'ast, 'tcx> MarkerConverterInner<'ast, 'tcx> {
         unsafe { transmute(rustc_span) }
     }
 
+    #[must_use]
+    pub fn to_resugared_span_id(&self, rustc_span: rustc_span::Span) -> SpanId {
+        if let Some(source_data) = rustc_span.source_callee() {
+            self.to_span_id(source_data.call_site)
+        } else {
+            panic!("driver requested resugared id for an unsugared span: {rustc_span:#?}");
+        }
```
Veetaha, line 82:
```suggestion
        let source_data = rustc_span.source_callee().unwrap_or_else(|| {
          panic!("driver requested resugared id for an unsugared span: {rustc_span:#?}");
        });

        self.to_span_id(source_data.call_site)
```

marker_uilints/tests/ui/expr/sugar/for_loop.stderr
```diff
@@ -0,0 +1,82 @@
...
+  = note: Snippet: "range"
+
+warning: expr
+  --> $DIR/for_loop.rs:9:20
+   |
+9  |       for i in range {
+   |  ____________________^
+10 | |         total += i;
+11 | |     }
+   | |_____^
```
Veetaha, line 43: I didn't think that the block of the `for` is a separate expression. Is this right?

marker_uilints/tests/ui/expr/sugar/while_loop.stderr
```diff
@@ -0,0 +1,96 @@
...
+note: the lint level is defined here
+  --> $DIR/while_loop.rs:7:12
+   |
+7  |     #[warn(marker::print_every_expr)]
+   |            ^^^^^^^^^^^^^^^^^^^^^^^^
+
+warning: expr
+ --> $DIR/while_loop.rs:8:11
+  |
+8 |     while let Some(_) = something {
```
Veetaha, line 25: It's also surprising that the entire `let pat = scrutinee` is considered to be an expression.

marker_uilints/tests/ui/print_cond_expr.stderr
```diff
@@ -1109,86 +1025,33 @@ warning: print test
                        id: ExprId(..),
                        span: SpanId(..),
                    },
-                   expr: Call(
-                       CallExpr {
+                   expr: Path(
```
Veetaha, line 1028: I suppose a bug/feature was fixed here somehow, because before it was a call expression?

marker_uilints/tests/ui/print_span.stderr
```diff
@@ -1,33 +1,50 @@
...
-                  "$DIR/print_span.rs",
-              ),
-              start: 134,
-              end: 137,
-          }
-  = note: Snippet: 178
+  = note: Debug: $DIR/print_span.rs:9:21 - 9:30
+  = note: Snippet: Some(21)?
+  = note: Source: File(
+              FileInfo {
```
Veetaha, line 10: Maybe for `FileInfo` the debug representation could be just the file path? This would reduce the snapshots removing the noisy `file:, span_src: SpanSrcId(..)` portions.

Veetaha (commented):
Nicely done 👍

marker_adapter/src/context.rs
```diff
@@ -7,7 +7,8 @@ use marker_api::{
     ast::{
         item::{Body, ItemKind},
         ty::SemTyKind,
-        BodyId, ExprId, ItemId, Span, SpanId, SymbolId, TyDefId,
+        BodyId, ExpnId, ExpnInfo, ExprId, FileInfo, FilePos, ItemId, Span, SpanId, SpanPos, SpanSource, SymbolId,
```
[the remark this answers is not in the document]
xFrednet: Rustc uses the *expn* abbreviation. I've tried looking for other Span implementations, but they all seem to only have the file layer.

There are a lot of names, that are bikeshedded in Marker. This is just another one in the collection ^^

marker_api/src/ast/common/id.rs
```diff
@@ -69,6 +69,11 @@ new_id! {
     pub GenericId: u64
 }
 
+new_id! {
+    /// This ID uniquely identifies a user defined type during linting.
```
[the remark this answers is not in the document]
xFrednet: Copy-pasta :spaghetti: 

Though, this refers to macros in general. I believe proc and derive macros also have unique IDs.

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
+/// }
+///
+/// ex4a!();
+/// ex4b!();
+/// ```
+///
+/// This example expands `ex4a` into a new `ex4b` macro, which in turn expands
+/// into a `4 + 4` expression. The [`Span`] of the binary expression has three
+/// layers. First the root layer calling the `ex4b` macro, this would then
+/// expand to the `4 + 4` expression, which actually comes from the `ex4a` expansion.
```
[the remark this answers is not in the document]
xFrednet: Nope, the layers should roughly be:
1. root
2. ex4a
3. ex4b
The second layer is added by the ex4a macro, not sure how to formulate it better

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
+///
+/// ### Proc Macros
+///
+/// Proc macros and some built-in macros are different from `macro_rules!` macros as
+/// they are opaque to the driver. It's just known that some tokens are provided as an
+/// input and somehow expanded. The [`Span`]s of the expanded tokens are marked as
+/// coming from an expansion by default. However, macro crates can sometimes override
+/// this with some trickery. (Please use this forbidden knowledge carefully.)
```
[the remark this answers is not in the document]
xFrednet: This is referring to `quote_spanned`, which can be a gigantic headache for linting. Clippy had several FPs and at times, just had to snip the source code to see if it matches with the AST. For linting reasons, I would either advise against using it at all, or making sure to always add `#[allow(clippy::all, clippy::pedantic)]`. This works with Clippy being the only linter, but it could again become a real problem for Marker

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
+pub struct ExpnInfo<'ast> {
+    _lifetime: PhantomData<&'ast ()>,
+    parent: ExpnId,
+    call_site: SpanId,
+    macro_id: MacroId,
+}
+
+impl<'ast> ExpnInfo<'ast> {
+    #[must_use]
+    pub fn parent(&self) -> Option<&ExpnInfo<'ast>> {
```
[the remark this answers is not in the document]
xFrednet: The root layer/source code is not an expansion. So the parent can be `None` if there is no further expansion. I'll add a coment about it.

For the `ExpnId`, that's only an internal ID, if it's a user-facing type, it should definitely be wrapped in an option. This is right now similar to rustc internals. They model the root level as a specific expansion. Wrapping it in an option would right now only waste memory and conversion code

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
+    /// let variable = 15_000;
+    /// //             ^^^^^^
+    /// //             lit_span
+    ///
+    /// lit_span.snippet(); // -> Some("15_000")
+    /// ```
+    ///
+    /// There are several reasons, why a snippet might be unavailable. Also
+    /// depend on the used driver. You can also checkout the other snippet
+    /// methods to better deal with these cases:
```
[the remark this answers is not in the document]
xFrednet: One example would be Rust-Analyzer, they wanted to have a stable AST representation, even before I started Marker. There is nothing really stopping them from also creating a driver. Complex lints should still use a compiler as a driver, but simple lints can in theory also be used in IDEs.

RA can also use Marker as a driver, like it can do for Clippy. But that would mean that the lint crates are reloaded every time. Personally, I also disabled that part of RA and always check in the console. 

Then it's just a general thing. If rustc was the only driver, it could have just wrapped pointers to rustc types. Looking back, that would have been better to have a minimal working prototype sooner, but this way there is a clear separation between the driver an API. Several internal types(mostly IDs) are based on rustc, but there is nothing stopping the development of another driver. 

cc: https://github.com/rust-marker/marker/issues/32

marker_api/src/ast/common/span.rs
xFrednet: I don't really see how the linked crate would improve this much.

The main problem I see with the current solution is the lifetime transmutes, but I don't see a way to store `&'ast AstContext<'ast>` in a static variable otherwise.

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
+    ///
+    /// For placeholders, it's recommended to use angle brackets with information
+    /// should be filled out. For example, if you want to snip an expression, you
+    /// should use `<expr>` as the default value.
+    ///
+    /// If you're planning to use this snippet in a suggestion, consider using
+    /// [`snippet_with_applicability`](Self::snippet_with_applicability) instead.
+    pub fn snippet_or(&self, default: &str) -> String {
+        self.snippet().unwrap_or(default).to_string()
+    }
```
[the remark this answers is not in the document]
xFrednet: > Or maybe this method could just not be provided?

What do you mean with this? Why shouldn't Marker provide this? All suggestions and diagnostic messages in rust are basically constructed using snippets. A user could use `snippet().unwrap_or()` instead, but that is longer and would also not have specific documentation.

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
+    /// For the placeholder, it's recommended to use angle brackets with information
+    /// should be filled out. A placeholder for an expression should look like
+    /// this: `<expr>`
+    ///
+    /// The applicability will never be upgraded by this method. When you draft
+    /// suggestions, you'll generally start with the highest [`Applicability`]
+    /// your suggestion should have, and then use it with this snippet function
+    /// to adjust it accordingly. The applicability is then used to submit the
+    /// suggestion to the driver.
+    pub fn snippet_with_applicability(&self, placeholder: &str, applicability: &mut Applicability) -> String {
```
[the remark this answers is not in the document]
xFrednet: Having the `&mut` parameter allows the user to use `snippet_with_app` inline, like this:

```rust,ignore
let mut app = Applicability::MachineApplicable;
let sugg = format!(
    "{}..{}",
    a.span().snippet_with_applicability("<expr-a>", &mut app),
    b.span().snippet_with_applicability("<expr-b>", &mut app),
);
```

I've added this example

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
+    /// Returns the length of the this [`Span`] in bytes.
+    pub fn len(&self) -> usize {
+        (self.start.0 - self.end.0)
+            .try_into()
+            .expect("Marker is not compiled for usize::BITs < 32")
     }
 
     /// Returns `true` if the span has a length of 0. This means that no bytes are
     /// inside the span.
     pub fn is_empty(&self) -> bool {
```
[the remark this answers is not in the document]
xFrednet: Empty spans can be useful for markers or to insert new code. Rustc sometimes provides a span, where something could be substituted. For example, when a user writes a function without a return type, but a lint want's to recommend adding one. Then there are no characters to point to and the span is empty.

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
-
-    pub fn start(&self) -> usize {
+    /// Returns the start position of this [`Span`].
+    pub fn start(&self) -> SpanPos {
         self.start
     }
 
-    pub fn set_start(&mut self, start: usize) {
+    /// Sets the start position of this [`Span`].
+    pub fn set_start(&mut self, start: SpanPos) {
```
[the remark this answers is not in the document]
xFrednet: Having functions to nicely combine spans is on the todo list I wanted to create a good-first-issue for it, once this PR is merged. Marker currently has very few issues, this can be picked up easily, that's why I haven't added them yet.

For the dangerous part, yeah, in Clippy we had a few panics due to that, but sometimes this is just needed to create better suggestions. I would rather not lock users out of creating great suggestions, especially if Marker might be used for automatic migrations. Marker can easily have a lint, to recommend users to use another method instead

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
+    }
+
+    pub fn source_id(&self) -> SpanSrcId {
+        self.source_id
+    }
+}
+
+#[repr(C)]
+#[derive(Debug)]
+#[allow(clippy::exhaustive_enums)]
```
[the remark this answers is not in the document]
xFrednet: This enum should actually be `[non_exhaustive]` I just pasted the wrong thing :sweat_smile: 

For `[non_exhaustive]`, Marker's goal is to create a stable and future-proof API. It can always happen that we need to add new variants to enums. That's also the reason I enabled the restriction lint. With application code, it's something different, since you also control all consumers of the code you write. Marker is just a library which might also be used for closed source lints. So API changes can not be taken lightly, after v1.0.0

marker_api/src/ast/common/span.rs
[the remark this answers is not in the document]
Veetaha: I see, so it's not easily possible to detect if the code comes from the proc macro expansion, and therefore the author of the proc macro should add an `#[allow(...)]` to work around that. That's a pity, we should definitely not rely on people adding `allow(...)` in their macros. Checking the source snippet for the presence of a macro may be better

marker_api/src/ast/common/span.rs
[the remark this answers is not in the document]
Veetaha: I guess listing them likes this in the comment will be fine.

marker_adapter/src/context.rs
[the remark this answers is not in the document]
Veetaha: Okay, I noticed it was used by rustc. I guess it's a good reason to stick with this name

marker_api/src/ast/common/span.rs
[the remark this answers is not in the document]
Veetaha: Yeah, documentation is a good point

marker_api/src/ast/common/span.rs
Veetaha: I think `scoped-tls` allows you to get rid of `&'ast` part, but it won't remove the `<'ast>` from the type, so you are right, it won't fix the transmutes.

marker_api/src/ast/common/span.rs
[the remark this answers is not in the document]
xFrednet: Jup, there is some discussion to maybe add more support in rustc, we'll see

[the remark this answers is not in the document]
xFrednet: > Checking the source snippet for the presence of a macro may be better

This is what Clippy currently does, but it sadly takes a significant amount of time.

marker_api/src/ast/common/span.rs
```diff
@@ -2,133 +2,469 @@ use std::marker::PhantomData;
...
+    }
+
+    /// Maps the given [`SpanPos`] to a [`FilePos`]. The [`SpanPos`] has to correspond
+    /// to a position that belongs to this [`FileInfo`].
+    pub fn to_file_pos(&self, span_pos: SpanPos) -> Option<FilePos> {
+        with_cx(self, |cx| cx.span_pos_to_file_loc(self, span_pos))
     }
 }
 
 #[cfg(feature = "driver-api")]
```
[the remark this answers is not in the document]
xFrednet: I assume that most users will just read the documentation on [docs.rs](https://docs.rs/marker_api/latest/marker_api/) which should not list these cfged item. Example: [`marker_api::ast::expr::FieldExpr`](https://docs.rs/marker_api/latest/marker_api/ast/expr/struct.FieldExpr.html) I considered moving them to a different module or even into `marker_adapter`.

I think this is the best solution for now. If we have the submodule in the same file, it wouldn't really improve anything. If we move it to a different file, we would need the struct fields to have a "higher" visibility.

marker_rustc_driver/src/context/storage.rs
```diff
@@ -1,21 +1,19 @@
...
 
 use bumpalo::Bump;
-use marker_api::ast::SpanSource;
-use rustc_hash::FxHashMap;
-
-use crate::conversion::common::SpanSourceInfo;
 
 pub struct Storage<'ast> {
+    /// The `'ast` lifetime is bound by the `buffer` field. Having it as a parameter
+    /// makes it easier to declare it in the return of functions
```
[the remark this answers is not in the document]
xFrednet: It's exactly for the second case, `&Storage` is stored in a few structs. Having the lifetime as an explicit parameter helps at least me to think about the scope. It also allows us to add cache fields, if needed in the future

marker_rustc_driver/src/conversion/marker/common/span.rs
xFrednet: I still think this should be a Clippy lint, as it's generally applicable

marker_rustc_driver/src/conversion/marker/common.rs
xFrednet: Here I don't really see the improvement of the suggesting. Having an if statement is clearer to me, and in this case, it doesn't save on indention.

marker_uilints/tests/ui/expr/sugar/for_loop.stderr
xFrednet: Yeah, that's how it's defined in the reference: [Iterator loops](https://doc.rust-lang.org/reference/expressions/loop-expr.html#iterator-loops)

It also makes the conversion and util code easier to reuse structs

marker_uilints/tests/ui/print_cond_expr.stderr
xFrednet: Yes, before `x?` was *resugared* to `TryTrait::branch(x)?` which was no correct. The removed path now corrects the resugar to `x?`

marker_uilints/tests/ui/expr/sugar/while_loop.stderr
xFrednet: The reference doesn't consider it a separate expression, but this causes a lot of headache. For example `while` and `while let` loops are officially considered different expressions, it's the same with `if` expressions then there is also the question how [if-let-chains](https://rust-lang.github.io/rfcs/2497-if-let-chains.html) should be modeled without that being its own expression.

In practice rustc and I believe RA both model it as an expression and Marker does as well. I think the main reason for now having it an official expression, is that you could otherwise write weird code like
```
for x in {let x = 17; [1, 2]} {
}
```

Return expressions are in this weird group, you can write them basically everywhere, like in this example:

```
    if return () {
        println!("Hey")
    }
```

marker_uilints/tests/ui/print_span.stderr
xFrednet: IDK, with rustc Span I often find it annoying, that they always print the path, since I often wanted to know the exact values they store. Reworking the `Debug` implementations of Marker structs to only include "public" information should be on the todo list, but not in this branch IMO

marker_rustc_driver/src/conversion/marker/common.rs
Veetaha: I used to think in early returns. The indentation is slightly improved, because the `self.to_span_id(...)` expression is now in closer to the left of the screen. It isn't very noticeable in this case, but in general I am used to think in this way:

1. Call a function
2. Process the error and early return or panic (consequently use `map_err(...)?` or `unwrap_or_else(|| panic!()` or `unwrap/expect`.
3. Proceed with the regular code flow on the same level of indentation.

And it helps a lot in functions where you have several functions where you process the error or panic. I suppose you didn't work with error/panic handling a lot (this is what your [comments](https://github.com/rust-marker/marker/blob/74554e7b151724c68995f91e2d6fa0207af7d8d0/cargo-marker/src/exit.rs#L1-L2) suggest, as well as some comments on other closed PRs that I saw).

From my experience using the consistent early-return approach with the minimum amount `if/else` helps keeping the code where every function call is fallible (and potentially async) much easier to read and maintain. I've written some conventions for my team, and this is one of them. You can see it [here](https://elastio.notion.site/Avoid-nesting-91ed9b81ce1f40bc9c3f268fed0445cf).

I know it's highly controversial just like the eternal argument between `to_owned/to_string`. The code in `marker_rustc_driver` is definitely not the type of code I am writing every day, so I'll leave such decisions to you. However, `cargo-marker` is much closer to my sphere, and I'd like to apply the best practices I know there in my PRs

marker_uilints/tests/ui/expr/sugar/while_loop.stderr
Veetaha: I see. I don't think that the let part of the `while/if let` syntax being an expression is a problem at least today. I was interested in the reason why that is.

Regarding the return expression I think it is not surprising that you can write them everywhere. The `return` is simply an operator just like an unary negation (`!`). Everywhere where you can write an unary negation (`!`) you should be able to write a return. It is not a special case in this regard.

marker_uilints/tests/ui/print_span.stderr
Veetaha: Beware that you can use the `{:#?}` syntax to print an alterante debug representation of your type. It isn't very flexible (it's just a boolean), but it allows you to customize the debug representation at least somehow.

This way if you use a terse form `{:?}` the file info could print just the file path, and if you use the `{:#?}` then the debug impl could include everything (the verbose debug output).

To implement this you need to use the [`alternate` method of the formatter](https://doc.rust-lang.org/stable/std/fmt/struct.Formatter.html#method.alternate)

marker_rustc_driver/src/conversion/marker/common/span.rs
Veetaha: Regarding the `to_owned/to_string`. I don't care too much about this, but it's better to use a consistent style anyway. I know there is a clippy lint for forcing the usage of [`to_owned`](https://rust-lang.github.io/rust-clippy/rust-1.56.0/index.html#str_to_string) over `to_string`. If you don't agree with that clippy lint, we can have a custom marker lint to force `to_string` in this repo. Anyway it's better to have some automation to force a consistent style for that and prevent the argument

marker_rustc_driver/src/conversion/marker/common.rs
xFrednet: > And it helps a lot in functions where basically every function call is fallible and you need to propagate the error or panic. I suppose you didn't work with error/panic handling a lot (this is what your [comments](https://github.com/rust-marker/marker/blob/74554e7b151724c68995f91e2d6fa0207af7d8d0/cargo-marker/src/exit.rs#L1-L2) suggest, as well as some comments on other closed PRs that I saw).
> 
> From my experience using the consistent early-return approach with the minimum amount of `if` and especially `else` helps keeping the code where every function call is fallible (and potentially async) much easier to read and maintain. I've written some conventions for my team, and this is one of them. You can see it [here](https://elastio.notion.site/Avoid-nesting-91ed9b81ce1f40bc9c3f268fed0445cf).

I agree that this is probably the main difference. I'm used to linting code, which is basically if's all the way down. For me, the `if` statement above parses quicker in my head than the `unwrap_or_else()` one. But in `cargo-marker` I've noticed the limitations of this style.

I guess it depends on how else cases are managed. In Clippy and lints you only care about the true case. With `if-let-chains` it's basically the same indention with fewer lines of code in comparison to early returns.

I'm totally fine with you using the other style. When working on `cargo-marker` I'll probably need to revise my style, anyways.

marker_uilints/tests/ui/print_span.stderr
xFrednet: Spans will most often be printed as part of other nodes. Therefore, I feel like the alternative print is not the best solution. I would like the API to this file output, whenever it would print `SpanId(...)`. That's a relatively simple change, but something for the future :)

Issue #129, opened 2023-05-03
Migrate to github merge queues


Niki4tap:
On May 1, 2023 [bors-ng](https://github.com/bors-ng/bors-ng) [became deprecated](https://bors.tech/newsletter/2023/05/01/tmib-76/).

> The main reason for this is that GitHub has a [built-in merge queue](https://github.blog/changelog/2023-02-08-pull-request-merge-queue-public-beta/) now that is publicly available... More importantly, there’s a bunch of bugs in bors-ng that we can’t fix, but they can.

Github merge queues seem overall better than bors-ng due to its integration with the service itself, and unfortunately the public instance [already throws warnings](https://github.com/bors-ng/bors-ng/commit/7430c3ad5b078d6c1d72c77e79a3a4ecbc28a3d2) and will eventually go down. This is acceptable for the projects that self-host a bors instance and cannot change their workflow, but I think in this project we can try new features.

More on merge queues [here](https://github.blog/changelog/2023-02-08-pull-request-merge-queue-public-beta/) and [here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue).

xFrednet:
In the bors-ng forum it was said, that bors will most likely continue to work, until GH's merges queues are stabilizes. This issue should definitely be addressed before then, but for now I'm focussing on other issues.

The main holdup on merging #132 is that the `bors try` comment would be lost. That one is super useful, since the normal CI only runs on linux, while the bors tests run on windows, linux, and macos.

Issue #221, opened 2023-08-19
Make it possible to use marker_uilints with cargo-nextest


Veetaha:
`cargo-nextest` is an awesome tool that considerably increases the tests performance. I won't describe all the problems that it solves and how it does that. This can be found in their [home page](https://nexte.st/index.html).

The problem is that `marker_uilints` doesn't support `cargo-nextest` right now. To be compatible with it this crate must implement the subset of `libtest` CLI API that [`cargo-nextest` requires](https://nexte.st/book/custom-test-harnesses.html).

xFrednet:
Clippy has some boilerplate to also support cargo-nextest:

https://github.com/rust-lang/rust-clippy/blob/7408f1d231a5e223a2ab01032ffd1778bcb6e0f1/tests/compile-test.rs#L326

I haven't used it yet, but this might be a good reference :)

PR #222, opened 2023-08-19
Replace bors with Github native merge queue

Veetaha:
## Configuration

You need to enable the merge queues manually in the repository settings like this.

<details>
<summary>Details</summary>

image
image
image

</details>

Be sure to specify all the required checks. Unfortuntelly they must be maintained separately in the repo settings. Maybe once this becomes a big enough problem and https://github.com/integrations/terraform-provider-github/issues/1481 is closed, we may use [`terraform`](https://registry.terraform.io/providers/integrations/github/latest/docs) to manage the github configurations as code.

It's possible to trigger a test workflow run using the Actions tab in github UI:

image

Not sure if this is in fact useful, but why not.

The merge queue can be observed under a special URL. That URL can be obtained elsewhere. For example, when you submit the PR into the queue, a link will appear where the merge button was. This is really stupid, but yeah, github didn't take care to create an explicit tab in the UI to job to the merge queue view.

<details>
<summary>Details</summary>

image

</details>

## Merging strategy

The `Squash and merge` method will produce a commit with the message that begins with the PR title, PR number, and the body of the message will contain all the commit messages combined together. There isn't a way to customize it, unfortunately. I wish github could put the PR description there (just like bors does that), but we are here to suffer.

image

Anyway, I don't think it is very important. The main thing is that the PR title is in the head of the commit message and the PR reference is also there, so if someone wants to read what the commit does they could read just the commit title, or follow the PR link for details.

## CI changes

Squashed the mdbook CI with into a shared `ci.yml`. The reason for that is that this way it is simpler. It also makes it possible to make the `mdbook` a required check in the repo config. Unfortunatelly, if we make this job conditional (running only on md files changes), then we won't be able to make it required in the repo config. I know another workaround that involves making each step of the job conditional, but that is so hacky and dirty, that I don't want to apply it at this stage. This job requires mere seconds to run, I don't think it makes sense to optimize it and make it conditional.

---

Split the rust checks, tests and lints into separate jobs. This way they all run in parallel and should be more efficient. With this change CI for all major platforms is going to run on PRs as well. I don't think it makes sense to run only `ubuntu` on CI. The CI run time is very small and we don't pay for it. Let's use all of its juice for early feedback.

---

Transitioned to `actions-rust-lang/setup-rust-toolchain@v1`. This action handles caching of `CARGO_HOME` and `./target` dir artifacts for us.

--- 

I tried to use `cargo-nextest` to speed up the test runs. But the problem is that `marker_uilints` has to support the interface of libtest, i.e. the CLI parameters required by `cargo-nextest` for custom harnesses: https://nexte.st/book/custom-test-harnesses.html. Maybe at some point, we'll fix this in https://github.com/rust-marker/marker/issues/221.

I even made the download scripts cross-platform to be able to download `cargo-nextest` on windows, but then found blocker described in #221. I left the changes to download scripts. At least the guys that use windows/mac will be able to use them to download various tools used in this repo on their platforms using the scripts from this repo

## Code changes

I removed the usage of `lint_reasons`	nightly feature from `marker_adapater` to make it compile on stable. I think it makes sense to use the stable toolchain and only stable features if possible. For example, `marker_lints` could also compile on stable, but I didn't tackle that as part of this PR.

Maybe `marker_rustc_driver` could live in a separate cargo workspace with the nightly toolchain in its `rust-toolchain` file. I didn't play with that, and not sure if it will compilate things instead.

---

Closes: https://github.com/rust-marker/marker/issues/129


xFrednet:
Thank you for the PR, I'll review it in the coming week :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,117 @@
...
+  rust-lint:
+    runs-on: ${{ matrix.os }}-latest
+
+    strategy:
+      matrix:
+        os: [ubuntu, windows, macos]
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
```
xFrednet, line 50: Where does `actions-rust-lang/setup-rust-toolchain@v1` get the toolchain from?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,117 @@
...
+
+      - run: cargo clippy
+      - run: cargo doc --no-deps
+
+  rust-formatting:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      # We don't need any cache here, so use dtolnay action directly
+      - uses: dtolnay/rust-toolchain@stable
```
xFrednet, line 60: Do we maybe want to pin the stable version as well, to ensure that they build on stable v1.64 or something like that? I think this would at least make sense for `cargo-marker`. Not a blocker for this PR, just a question :)

xFrednet (approved):

xFrednet:
LGTM, just two questions, and then I can enable merge queues and merge this :)

.github/workflows/ci.yml
Veetaha: It reads it from the `rust-toolchain.toml` file

.github/workflows/ci.yml
Veetaha: I also thought that it makes sense to pin the toolchain version. I think it makes sense to select the latest stable version target for `cargo-marker` once we add pre-compiled binaries to the release process.

Veetaha:
Sorry I had to force-push because I just noticed I had local changes that were done with `git commit --amend` but were not pushed.

The diff is this

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 33f351b..b105a6c 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,12 @@ defaults:
 
 env:
   RUST_BACKTRACE: 1
-  CARGO_INCREMENTAL: 0
   CARGO_TERM_COLOR: always
   RUSTDOCFLAGS: --deny warnings
   RUSTFLAGS: --deny warnings
+  # This disables incremental compilation for workspace packages and path deps.
+  # All other dependencies including registry deps will still use the incremental cache.
+  CARGO_INCREMENTAL: 0
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -57,7 +59,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       # We don't need any cache here, so use dtolnay action directly
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@1.71.1
       - run: cargo fmt --check
 
   # This job ensures required packages can be built with a stable toolchain
@@ -74,7 +76,7 @@ jobs:
       - run: rm rust-toolchain.toml
       - uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          toolchain: stable
+          toolchain: 1.65.0
 
       - run: >-
           cargo check
```

Veetaha:
Hey @xFrednet I see that CI now fails on [this error](https://github.com/rust-marker/marker/actions/runs/5919054664/job/16048948762?pr=222#step:5:104):

```
error: `-Csplit-debuginfo=unpacked` is unstable on this platform
```

I suppose support for unpacked debug info on windows was added in some recent version of rust toolchain? Could you elaborate why this setting was enabled in `.cargo/config.toml`?

xFrednet:
> I suppose support for unpacked debug info on windows was added in some recent version of rust toolchain? Could you elaborate why this setting was enabled in .cargo/config.toml?

We use that setting in Clippy, as it slightly speeds up compile times on some platforms, if it makes trouble, you can remove it again

xFrednet:
> Sorry I had to force-push because I just noticed I had local changes that were done with git commit --amend but were not pushed.

No need to be sorry :) Simple force pushes diffs are also displayed correctly by GH

Veetaha:
I wouldn't like to delete the config file in CI because it may contain additional configs in the future. I'll just disable this for now.

Veetaha:
Looks like CI is green now. So it can probably be merged

xFrednet (approved):

xFrednet:
This version looks good to me, thank you very much! I'll merge this later today, when I have a bit more time to read through all the settings again and debug if necessary.

xFrednet:
> Merging strategy

I like having the option to squash, but it seems like GH requires us to select a single mode. I'll let it be the default merge commit one for now. I don't want to lose the commit structure or PRs, especially if they're still affecting many areas right now.

Let's see how this merge will go :)

xFrednet:
Well, the queue seems to be working, but the book deployment broke, I'll see if I can fix it

PR #223, opened 2023-08-20
Chore: Add copyright notice

xFrednet:


Veetaha (commented):
I think it should be possible to create symlinks to the copyright file in individual crates. 
Serde, for example, [does this](https://github.com/serde-rs/serde/tree/master/serde_derive)

xFrednet:
Good point, thank you for the recommendation, that's 100 times better :heart:. I also verified that it gets packaged correctly :)

bors r=Veetaha

PR #224, opened 2023-08-21
CI: Fix Marker Book Deployment

xFrednet:
Fix book deployment after switching to merge queues in rust-marker/marker#222

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -36,7 +36,7 @@ jobs:
           mkdir deploy
           cp ./docs/index.html ./deploy/index.html
       - name: Build the book
-        run: mdbook build docs/book --dest-dir deploy/book
+        run: mdbook build docs/book --dest-dir ../../deploy/book
 
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v2


.github/workflows/gh-pages.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           mkdir deploy
           cp ./docs/index.html ./deploy/index.html
       - name: Build the book
-        run: mdbook build docs/book --dest-dir deploy/book
+        run: mdbook build docs/book --dest-dir ../../deploy/book
```
Veetaha, line 39: That's strange. The step above puts the `index.html` into `./deploy`. The why is this `../../deploy`?

Veetaha: Ah, I see this is in the help:
```
  -d, --dest-dir <dest-dir>  Output directory for the book
                             Relative paths are interpreted relative to the book's root directory.
                             If omitted, mdBook uses build.build-dir from book.toml or defaults to `./book`.
```

xFrednet: The path after `mdbook build` changes the directory that `mdbook build` is executed. The `--det-dir` as to be relative to the book source directory. It's a weird thing, but somehow works (hopefully)

xFrednet: I ran this locally and it put everything under `./deploy/book` afterwards. The `book` sub dir is to have it as a sub url

xFrednet:
Now everything works :tada: Thank you updating the CI @Veetaha :heart:

Issue #226, opened 2023-08-22
Use the `workspace.package` table


xFrednet:
Cargo allows us to define dependencies and some package information on the workspace level and have packages inherit them. I want to use this for Marker as well :)

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

PR #227, opened 2023-08-22
Chore: Use `workspace.package` and `workspace.dependencies` and bump version to `v0.2.0-dev`

xFrednet:
I hope this is the correct usage of `workspace` information. I've bumped the version in the manifest, but left the version in the documentation as `0.1.1`, since I believe that new comers will first look at the repository and should therefore see the last released version and not the name of the new one. I'll maybe write a script at some point, to automatically update all versions. Declaring the version on the workspace level is already a major improvement :)

---

Closes: rust-marker/marker#226

### DIFF
--- a/cargo-marker/Cargo.toml
+++ b/cargo-marker/Cargo.toml
@@ -1,13 +1,14 @@
 [package]
-name    = "cargo_marker"
-version = "0.1.1"
+name = "cargo_marker"
 
 categories  = ["development-tools::cargo-plugins"]
 description = "Marker's CLI interface to automatically compile and run lint crates"
-edition     = "2021"
 keywords    = ["marker", "linting", "cli", "cargo", "cargo-subcommand"]
-license     = "MIT OR Apache-2.0"
-repository  = "https://github.com/rust-marker/marker"
+
+edition    = { workspace = true }
+license    = { workspace = true }
+repository = { workspace = true }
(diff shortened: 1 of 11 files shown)


cargo-marker/Cargo.toml
```diff
@@ -17,10 +18,10 @@ name = "cargo-marker"
...
 
 [dependencies]
-camino         = { version = "1.1", features = ["serde1"] }
-cargo_metadata = "0.15.4"
-clap           = { version = "4.0", features = ["string", "derive"] }
-once_cell      = "1.17.0"
-serde          = { version = "1.0", features = ["derive"] }
-serde_json     = "1.0"
-toml           = "0.7"
+camino.workspace         = true
```
Veetaha, line 21: I suggest using `key = { worspace = true }` instead.
This way it will align nicer.

Compare this code in the diff with this:

```toml
camino         =  { workspace = true }
cargo_metadata =  { workspace = true }
clap           =  { workspace = true }
once_cell      =  { workspace = true }
serde          =  { workspace = true }
serde_json     =  { workspace = true }
toml           =  { workspace = true }
```
Everything is now aligned with `taplo`, and the word `.workspace` doesn't clutter the package name

xFrednet:
Sounds good to me, I'll update tomorrow before I do the release :)

Issue #219, opened 2023-08-18
Nightly bump 2023-08-24


xFrednet:
Currently, Marker updates the nightly version on the day of the release, before releasing the next marker version. I find this a bit risky, as we have basically no time to test the nightly version with Marker. This release, I want to first release marker and then update the used nightly version. Let's see how that'll go.

A few PR's that might effect the sync:
* https://github.com/rust-lang/rust/pull/114781 
* https://github.com/rust-lang/rust/pull/114772

PR #230, opened 2023-08-24
Bump nightly -> 2023-08-24

xFrednet:
Another showcase, why we need something like Marker. This PR migrates every lint crate that uses marker's public API to use `nightly-2024-08-24` :)

Closes: https://github.com/rust-marker/marker/issues/219


xFrednet:
Forgot to check for new Clippy lints, now it's ready :)

PR #233, opened 2023-09-01
Type constraint for paths and IntoUtf8 trait

nitkach:
## UTF-8 encoding trait `IntoUtf8`
Trait supports conversion `Vec<u8> -> String` and `PathBuf -> Utf8PathBuf`. 

## `Utf8Path`
Changing the path type to `Utf8Path` and `Utf8PathBuf` ensures that they have only UTF-8 encoding.
And also, they implement `Display`, which is also very convenient :)

### DIFF
--- a/cargo-marker/Cargo.toml
+++ b/cargo-marker/Cargo.toml
@@ -36,4 +36,4 @@ tracing-subscriber = { workspace = true, features = ["env-filter"] }
 yansi              = { workspace = true }
 
 [dev-dependencies]
-expect-test = "1.4"
+expect-test = { workspace = true }
--- a/cargo-marker/Cargo.toml
+++ b/cargo-marker/Cargo.toml
@@ -34,3 +34,6 @@ tracing            = { workspace = true }
 tracing-error      = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 yansi              = { workspace = true }
+
+[dev-dependencies]
+expect-test = "1.4"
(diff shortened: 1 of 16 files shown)


cargo-marker/Cargo.toml
```diff
@@ -34,3 +34,6 @@ tracing            = { workspace = true }
 tracing-error      = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 yansi              = { workspace = true }
+
+[dev-dependencies]
+expect-test = "1.4"
```
Veetaha, line 39: Let's put this version requirement in the `[workspace,dependencies]` in the root-level `Cargo.toml`

cargo-marker/src/backend/toolchain.rs
```diff
@@ -158,10 +158,10 @@ pub(crate) fn rustup_which(toolchain: &str, tool: &str) -> Result<PathBuf> {
         return Err(Error::wrap(stderr.trim(), format!("Command failed: {}", cmd.display())));
     }
 
-    let string_path = String::from_utf8(output.stdout).context(|| "Incorrect bytes")?;
-    let path = PathBuf::from(string_path.trim());
+    let string_path = output.stdout.into_utf8()?;
+    let path = Utf8PathBuf::from(string_path.trim());
 
-    info!(%tool, %toolchain, path = %path.display(), "Found the tool");
+    info!(%tool, %toolchain, path = %path, "Found the tool");
```
Veetaha, line 164:
```suggestion
    info!(%tool, %toolchain, %path, "Found the tool");
```

marker_rustc_driver/src/main.rs
```diff
@@ -403,3 +382,17 @@ impl From<rustc_span::ErrorGuaranteed> for MainError {
         Self::Rustc(err)
     }
 }
+
+#[test]
```
Veetaha, line 386: It's conventional to put tests under a module at the end of the file like this
```rust
#[cfg(test)
mod tests {
    #[test]
    fn my_test() {}
}
```

Veetaha (approved):

Issue #240, opened 2023-09-11
[Discussion] Review accepted RFCs and their implication/effects on Marker's AST 


xFrednet:
Every once in a while, I see a RFC that affects the grammar of Rust and with that probably also the AST representation that Marker provides. Before `v1.0.0`, we should review all accepted RFCs and discuss their implications for Marker and how these changes should be handled in the future.

For now, I'll add RFCs whenever I see them. This is not an exhaustive list:
* https://github.com/rust-lang/rfcs/pull/2102

PR #241, opened 2023-09-11
API: Rename `QuestionMarkExpr` -> `TryExpr` (Breaking change)

xFrednet:
The `?` operator is officially called the *ErrorPropagationExpression* in [Rust's grammar](https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-question-mark-operator). However, I've rarely seen people actually call it by that name. Even the documentation doesn't call it that after the grammar definition.

Before, I've called it the *QuestionMarkOperator*, but this name feels a bit off, as that describes how it's written, but not what it does. I have the impression, that the community commonly calls this operator the *Try* operator. That's at least the name I would use and what I've seen in a few blog posts.

This change is open for discussion. If you have any thoughts, please share them! :)

### DIFF
--- a/marker_api/src/ast/expr.rs
+++ b/marker_api/src/ast/expr.rs
@@ -61,7 +61,7 @@ pub enum ExprKind<'ast> {
     UnaryOp(&'ast UnaryOpExpr<'ast>),
     Ref(&'ast RefExpr<'ast>),
     BinaryOp(&'ast BinaryOpExpr<'ast>),
-    QuestionMark(&'ast QuestionMarkExpr<'ast>),
+    QuestionMark(&'ast TryExpr<'ast>),
     Assign(&'ast AssignExpr<'ast>),
     As(&'ast AsExpr<'ast>),
     Path(&'ast PathExpr<'ast>),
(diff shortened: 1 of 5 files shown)


jhpratt:
Given that there is an unstable `Try` trait that it desugared to, that name seems reasonable.

marker_api/src/ast/expr.rs
```diff
@@ -61,7 +61,7 @@ pub enum ExprKind<'ast> {
     UnaryOp(&'ast UnaryOpExpr<'ast>),
     Ref(&'ast RefExpr<'ast>),
     BinaryOp(&'ast BinaryOpExpr<'ast>),
-    QuestionMark(&'ast QuestionMarkExpr<'ast>),
+    QuestionMark(&'ast TryExpr<'ast>),
```
Veetaha, line 64: I suppose the variant needs to be renamed as well. But.. how should it be called? Just `Try`? That may be confusing with the try blocks. I would be fine with this renaming if it wasn't for the try blocks. It may be very confusing when try blocks become stable to have a TryExpr, which isn't a try block

xFrednet: Good catch! For variant name, I would currently call it `Try`. When the `TryBlock` is added, I would first look, if it can be merged with the `BlockExpr` and if not, add a new expression node, called `TryExpr`, here. That's also how syn, calls these nodes: https://docs.rs/syn/latest/syn/enum.Expr.html#variant.Try

xFrednet: I've added the RFC to #240

Veetaha: Okay, I think it should be good to be consistent with `syn`. If `syn` calls this a `TryExpr`, I'm fine with this change

Issue #242, opened 2023-09-11
[Feat]: Add a method to get the parent node of a given node


xFrednet:
### Summary

Requesting the parent of a given node, is a fundamental feature required to build context-sensitive lints. It's useful to check the context that an expression occurs in or to adapt the suggestion when required. Rustc provides a [`find_parent(HirId)`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/hir/map/struct.Map.html#method.find_parent) function and an iterator function to walk through the parents.

I've already planned a bit for this issue, and it's surprisingly complex. It might take some time to implement it properly and feedback from the community.

xFrednet:
The new addition of the `NodeId` struct can be used for the development of this issue.

I still have a tiny hope that I can work on this before v0.3.0, but it's likely that this will be moved to v0.4.0 as major breaking changes are currently taking priority for me.

xFrednet:
I did some work in preparation for this issue, but it's unlikely that I'll get this done before v0.3. It's free for the taking :)

Veetaha:
I've been thinking of implementing iterator API for traversing the AST tree. There is a major problem with implementing such API in that the iterator's state needs to have a dynamically allocated data structure (e.g. a stack) to track the position of the cursor inside of the AST. For example, each time the iteration enters a subtree the stack needs to grow and each time it exists the stack shrinks.

This dynamic state is conceptually unavoidable because the cursor needs to know where to return (how to go up in the tree). If there are parent node reference it would reduce the number of state needed to implement the iterator API, but still not entirely solve the problem. When we go up in the tree we also need to know where to continue (the index of the child node that was exited in the parent node plus one).

xFrednet:
We could also consider storing references of the parents in the nodes.

For the iterator, I think it depends on the flexibility we want to provide. If we only allow the traversal upwards, we only need a single ID in the iterator, that is used to request the next parent. If we allow going down, then it will be more complicated, like you described.

In Clippy, we usually only traverse upwards, if needed we might use a visitor to go downwards. I'm not sure if there is a lint, which would require an iterator that is traversable in both directions.

PR #244, opened 2023-09-15
API: Wrap `*Kind` enums in `*Kind` enums

xFrednet:
The current API has a few spots where a `*Kind` enum directly wraps a `*Kind` enum, like: [`PatKind::Lit`](https://docs.rs/marker_api/latest/marker_api/ast/pat/enum.PatKind.html#variant.Lit). This is cool, for matching but causes a bunch of problems when it comes to common data shared among the outer `*Kind` enum. To be future-proof and have this consistent in the entire API, I've now wrapped the inner `*Kind` values in new structs.

These structs are often just wrappers, but will help us in the future. This will help me a lot with https://github.com/rust-marker/marker/issues/242

Most of the changes are just uilint tests.


marker_api/src/ast/stmt.rs
```diff
@@ -1,24 +1,37 @@
...
+    /// Returns the [`Span`] of this statement.
+    fn span(&self) -> &Span<'ast>;
+}
 
 #[repr(C)]
 #[non_exhaustive]
 #[derive(Debug, Copy, Clone)]
 pub enum StmtKind<'ast> {
-    Item(&'ast ItemKind<'ast>, CtorBlocker),
+    Item(&'ast ItemStmt<'ast>),
```
Veetaha, line 24: Did you consider following the naming used in `syn` where you can see a pattern [like this](https://docs.rs/syn/2.0.33/syn/enum.Expr.html#variants):
```rust
enum Expr {
    Array(ExprArray),
    Assign(ExprAssign),
    // ...
}
```
However, a bit different naming should be used if some of the types may be reused in different contexts. For example, items not always have to appear inside of statements, os they shouldn't be called `ItemStmt` or `StmtItem`, but just `Item`. `Let` bindings may occur as part of `if/while let` , so maybe that part should also be just `Let`.
```rust
enum Stmt {
    Let(Let),
    Item(Item),
}
```
I'm definitely is wrong in this thinking because "let statement" also includes a semicolon at the end...  I haven't yet fully grasped the naming patterns in `marker_api`, so correct me if I'm wrong and this code uses an already-established extensible and workable naming scheme

marker_api/src/ast/stmt.rs
```diff
@@ -1,24 +1,37 @@
...
-};
+use crate::{ffi::FfiOption, private::Sealed};
+
+use super::{expr::ExprKind, item::ItemKind, pat::PatKind, ty::SynTyKind, Span, SpanId, StmtId};
+
+/// This trait combines methods, which all statements have in common.
+///
+/// This trait is only meant to be implemented inside this crate. The `Sealed`
+/// super trait prevents external implementations.
+pub trait StmtData<'ast>: Debug + Sealed {
```
Veetaha, line 12: Should this and a bunch of other traits be a `prelude` of `marker_api`?

marker_api/src/ast/stmt.rs
```diff
@@ -79,23 +132,67 @@ impl<'ast> LetStmt<'ast> {
...
+}
+
+impl_stmt_data!(ItemStmt<'ast>, Item);
+
+#[cfg(feature = "driver-api")]
+impl<'ast> ItemStmt<'ast> {
+    pub fn new(data: CommonStmtData<'ast>, item: ItemKind<'ast>) -> Self {
+        Self { data, item }
+    }
+}
```
Veetaha, line 198: Could it make sense to just mark the fields of the structs conditionally `pub` instead?

marker_rustc_driver/src/conversion/marker/stmts.rs
```diff
@@ -1,37 +1,44 @@
...
                     local.els.map(|els| self.to_expr_from_block(els)),
-                )))),
+                ))),
                 hir::LocalSource::AssignDesugar(_) => {
                     unreachable!("this will be handled by the block expr wrapping the desugar")
                 },
                 hir::LocalSource::AsyncFn | hir::LocalSource::AwaitDesugar => {
                     eprintln!("skipping not implemented statement at: {:?}", stmt.span);
-                    None
+                    None?
```
Veetaha, line 29: That's quite a hacky way to just say this:
```suggestion
                    return None;
```
The problem with `None?` is that `rustc`'s code flow analysis doesn't infer that this line of code returns, and thus won't infer an uninhabited (`!`) type of the expression here which may be important in some cases.

marker_rustc_driver/src/conversion/marker/pat.rs
```diff
@@ -41,7 +38,7 @@ impl<'ast, 'tcx> MarkerConverterInner<'ast, 'tcx> {
                 let lhs = lhs_map.get(id);
                 #[allow(clippy::unnecessary_unwrap, reason = "if let sadly breaks rustfmt")]
                 if pat.is_none() && matches!(mutab, rustc_ast::Mutability::Not) && lhs.is_some() {
-                    PatKind::Place(*lhs.unwrap(), CtorBlocker::new())
```
Veetaha, line 44: Could we get rid `CtorBlocker` type in this PR? That seems like an uncommon hack

marker_api/src/ast/stmt.rs
xFrednet: Yes, should be :+1:

marker_api/src/ast/stmt.rs
xFrednet: The `visibility` crate sadly doesn't work on field structs. Otherwise, that could be an option. I considered adding something like [`impl_new`](https://crates.io/crates/impl_new) to cfg derive these things. I was a bit hesitant with Marker's FFI types, but I just saw that it uses `impl Into<Ty>` that could actually work :thinking:

marker_rustc_driver/src/conversion/marker/pat.rs
xFrednet: It is unusual, but I sadly don't see a better option. If we take [`LitExprKind::UnaryOp`] for example. We want to make sure that the variant is only filled with a `-` operator, when it's used as a sign on a literal. Otherwise, it would not be a `LitExprKind`. Without the blocker, we would always need to make sure that the kind wasn't constructed by the user and nests some unexpected nodes.

Ideally, I would have liked to use `#[non_exhaustive]` but that only works on variants with fields. Tuple variants are made private, with that attribute. We could use variants with named fields, but I generally like tuple variants more for matching. I find them more readable and less verbose.

I haven't found or seen a better hack yet, and this one seems to work pretty well IMO, even if it is a hack

marker_api/src/ast/stmt.rs
xFrednet: Interesting that `syn` uses a similar AST structure, which I would have found that early on. :)

Marker's AST structure usually has a `*Kind` enum at the top level. This was inspired from rustc, which also uses `*Kind` enum, in nodes (See [`Expr`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/struct.Expr.html) with the `.kind` fiel of type [`ExprKind`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/enum.ExprKind.html)). It was called `Expr` for some time, but that felt a bit off to me for some reason. Guess it felt like the enums is not the actual expression, since it doesn't hold any data and just the different variant structs do.

The variant structs always have the type at the end. So any enum wrapped by `ExprKind` should end with `Expr`. This means an `if` expression is called `IfExpr`. Having the type in the end makes it more natural to read IMO. If you compare these two:

```
// Reads to me as "a takes a parameter `x` with the type `ExprIf`"
fn a(x: &ExprIf<'ast>) {}

// Reads to me as "a takes an if expression as parameter `x`"
fn a(x: &IfExpr<'ast>) {}
```

When I started the AST representation, I tried to find some examples, but didn't find any (And apparently ignored `syn` at that time). So this naming convention is just a result of what felt right and what I'm used to from rustc and [javaparser](https://javadoc.io/doc/com.github.javaparser/javaparser-core/latest/index.html) which I also used in a different project beforehand.

marker_api/src/ast/stmt.rs
Veetaha: I wouldn't add a proc macro dependency to `marker_api`. I love the fact that it can be a zero dependency crate, meaning the lint crates can also be effectively zero-dep (except for `marker_api`), and compile blazingly fast

xFrednet: The macro would only be needed when it's compiled with the `driver-api` feature, so it would remain zero-dep :D

Veetaha: Ah, right!

xFrednet: I'm undecided on this one. [impl_new](https://crates.io/crates/impl_new) looks great but only has a total of 102 downloads. [derive-new](https://crates.io/crates/derive-new) is from `@nrc` but doesn't seem to support `Into<X>` parameters, which is important for marker. It also doesn't seem like the crate isn't actively maintained, when I look at the [issue tracker](https://github.com/nrc/derive-new/issues) :thinking:

Veetaha: I'd recommend you to use either [`typed-builder`](https://github.com/idanarye/rust-typed-builder), for more elaborate cases: [`buildstructor`](https://docs.rs/buildstructor/latest/buildstructor/)

Veetaha (approved):

marker_api/src/ast/stmt.rs
xFrednet: I'll try using `typed-builder` instead of the newly added constructors, wish me luck!

xFrednet: It seems to work quite well. I won't update everything rn, but I'll keep the dependency and use it for new types :)

Here are the changes: https://github.com/rust-marker/marker/pull/244/commits/367cfb5680ed43f597b8b6edce64a08d06ad6c31

Veetaha (approved):

marker_api/src/ast/stmt.rs
xFrednet: There was also another lint that caused problems `clippy::used_underscore_binding` (https://github.com/idanarye/rust-typed-builder/issues/113)

Let's hope everything works now :)

xFrednet:
Thank you for the review @Veetaha, I highly appreciate it :heart:

PR #245, opened 2023-09-16
API: Reduce parameters for `AstContext::emit_lint` and add docs

xFrednet:
This came up in the discussion of rust-marker/marker#189. I believe the previous design was the correct technical implementation, I agree that it's quite verbose and potentially confusing to new users. Now we have a bunch of documentation, how `AstContext::emit_lint` should be used, and it only takes three parameters for a minimal lint emission. :tada: 

@Veetaha would you mind providing feedback on this PR? Since this came up in the discussion from rust-marker/marker#189

### DIFF
--- a/marker_api/src/ast/common/id.rs
+++ b/marker_api/src/ast/common/id.rs
@@ -39,6 +39,8 @@ macro_rules! new_id {
 
 use new_id;
 
+use crate::private::Sealed;
+
 new_id!(
     /// This ID uniquely identifies a crate during linting.
     pub CrateId: u32
@@ -144,3 +146,63 @@ new_id! {
     /// This ID uniquely identifies a statement during linting.
     pub StmtId: u64
 }
+
+#[repr(C)]
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy)]
+pub enum NodeId {
+    Expr(ExprId),
+    Item(ItemId),
+    Stmt(StmtId),
+    Body(BodyId),
+    Field(FieldId),
+    Variant(VariantId),
+}
+
+macro_rules! impl_into_node_id_for {
+    ($variant:ident, $ty:ty) => {
+        impl From<$ty> for NodeId {
+            fn from(value: $ty) -> Self {
+                NodeId::$variant(value)
+            }
+        }
+
+        impl From<&$ty> for NodeId {
+            fn from(value: &$ty) -> Self {
+                NodeId::$variant(*value)
+            }
+        }
+    };
+}
+
+impl_into_node_id_for!(Expr, ExprId);
+impl_into_node_id_for!(Item, ItemId);
+impl_into_node_id_for!(Stmt, StmtId);
+impl_into_node_id_for!(Body, BodyId);
+impl_into_node_id_for!(Field, FieldId);
+impl_into_node_id_for!(Variant, VariantId);
+
+pub trait HasNodeId: Sealed {
+    /// Returns the [`NodeId`] of the identifiable node
+    fn node_id(&self) -> NodeId;
+}
+
+impl<N: HasNodeId> HasNodeId for &N {
+    fn node_id(&self) -> NodeId {
+        (*self).node_id()
+    }
+}
+
+macro_rules! impl_identifiable_for {
+    ($ty:ty$(, use $data_trait:path)?) => {
+        impl<'ast> $crate::ast::HasNodeId for $ty {
+            fn node_id(&self) -> $crate::ast::NodeId {
+                $(
+                    use $data_trait;
+                )*
+                self.id().into()
+            }
+        }
+    };
+}
+pub(crate) use impl_identifiable_for;
--- a/marker_api/src/ast/common/span.rs
+++ b/marker_api/src/ast/common/span.rs
@@ -1,6 +1,6 @@
 use std::marker::PhantomData;
 
-use crate::{context::with_cx, diagnostic::Applicability, ffi};
+use crate::{context::with_cx, diagnostic::Applicability, ffi, private::Sealed};
 
 use super::{ExpnId, MacroId, SpanId, SpanSrcId, SymbolId};
 
@@ -386,6 +386,13 @@ impl<'ast> Span<'ast> {
     }
 }
 
+impl<'ast> HasSpan<'ast> for Span<'ast> {
+    fn span(&self) -> &Span<'ast> {
+        self
+    }
+}
+impl Sealed for Span<'_> {}
+
 #[cfg(feature = "driver-api")]
 impl<'ast> Span<'ast> {
     #[must_use]
@@ -509,11 +516,14 @@ impl<'ast> Ident<'ast> {
     pub fn name(&self) -> &str {
         with_cx(self, |cx| cx.symbol_str(self.sym))
     }
+}
 
-    pub fn span(&self) -> &Span<'ast> {
+impl<'ast> HasSpan<'ast> for Ident<'ast> {
+    fn span(&self) -> &Span<'ast> {
         with_cx(self, |cx| cx.span(self.span))
     }
 }
+impl<'ast> crate::private::Sealed for Ident<'ast> {}
 
 #[cfg(feature = "driver-api")]
 impl<'ast> Ident<'ast> {
@@ -570,3 +580,28 @@ impl_ident_eq_for!(
     std::ffi::OsString,
     std::borrow::Cow<'_, str>
 );
+
+/// A trait for nodes, that provide a [`Span`].
+pub trait HasSpan<'ast>: Sealed {
+    /// This returns the [`Span`] of the implementing AST node.
+    fn span(&self) -> &Span<'ast>;
+}
+
+/// This macro implements the [`HasSpan`] trait for data types, that provide a
+/// `span()` method.
+macro_rules! impl_spanned_for {
+    ($ty:ty) => {
+        impl<'ast> $crate::ast::HasSpan<'ast> for $ty {
+            fn span(&self) -> &$crate::ast::Span<'ast> {
+                self.span()
+            }
+        }
+    };
+}
+pub(crate) use impl_spanned_for;
+
+impl<'ast, N: HasSpan<'ast>> HasSpan<'ast> for &N {
+    fn span(&self) -> &Span<'ast> {
+        (*self).span()
+    }
+}
--- a/marker_api/src/context.rs
+++ b/marker_api/src/context.rs
@@ -11,7 +11,7 @@ use crate::{
         BodyId, ExpnId, ExpnInfo, ExprId, FileInfo, FilePos, ItemId, Span, SpanId, SpanPos, SpanSource, SymbolId,
         TyDefId,
     },
-    diagnostic::{Diagnostic, DiagnosticBuilder, EmissionNode},
+    diagnostic::{Diagnostic, DiagnosticBuilder, EmissionNode, EmissionNodeId},
     ffi,
     lint::{Level, Lint, MacroReport},
 };
@@ -93,17 +93,6 @@ impl<'ast> std::fmt::Debug for AstContext<'ast> {
     }
 }
 
-impl<'ast> std::hash::Hash for AstContext<'ast> {
-    fn hash<H: std::hash::Hasher>(&self, _state: &mut H) {}
-}
-
-impl<'ast> std::cmp::PartialEq for AstContext<'ast> {
-    fn eq(&self, other: &Self) -> bool {
-        std::ptr::eq(self, other)
-    }
-}
-impl<'ast> std::cmp::Eq for AstContext<'ast> {}
-
 #[cfg(feature = "driver-api")]
 impl<'ast> AstContext<'ast> {
     pub fn new(driver: &'ast DriverCallbacks<'ast>) -> Self {
@@ -112,27 +101,25 @@ impl<'ast> AstContext<'ast> {
 }
 
 impl<'ast> AstContext<'ast> {
-    pub fn lint_level_at(&self, lint: &'static Lint, node: impl Into<EmissionNode>) -> Level {
-        self.driver.call_lint_level_at(lint, node.into())
+    pub fn lint_level_at(&self, lint: &'static Lint, node: impl EmissionNode<'ast>) -> Level {
+        self.driver.call_lint_level_at(lint, node.emission_node_id())
     }
 
     #[allow(clippy::needless_pass_by_value)] // `&impl ToString`
-    pub fn emit_lint<F>(
-        &self,
-        lint: &'static Lint,
-        node: impl Into<EmissionNode>,
-        msg: impl ToString,
-        span: &Span<'ast>,
-        decorate: F,
-    ) where
+    pub fn emit_lint<F>(&self, lint: &'static Lint, node: impl EmissionNode<'ast>, msg: impl ToString, decorate: F)
+    where
         F: FnOnce(&mut DiagnosticBuilder<'ast>),
     {
+        let Some(span) = node.emission_span(self) else {
+            // If the `Span` is none, we can't emit a diagnostic message for it.
+            return;
+        };
         if matches!(lint.report_in_macro, MacroReport::No) && span.is_from_expansion() {
             return;
         }
-        let node = node.into();
+        let id = node.emission_node_id();
         if self.lint_level_at(lint, node) != Level::Allow {
-            let mut builder = DiagnosticBuilder::new(lint, node, msg.to_string(), span.clone());
+            let mut builder = DiagnosticBuilder::new(lint, id, msg.to_string(), span);
             decorate(&mut builder);
             builder.emit(self);
         }
@@ -247,7 +234,7 @@ struct DriverCallbacks<'ast> {
     // can't call them in safe Rust passing a &() pointer. This will trigger UB.
 
     // Lint emission and information
-    pub lint_level_at: extern "C" fn(&'ast (), &'static Lint, EmissionNode) -> Level,
+    pub lint_level_at: extern "C" fn(&'ast (), &'static Lint, EmissionNodeId) -> Level,
     pub emit_diag: for<'a> extern "C" fn(&'ast (), &'a Diagnostic<'a, 'ast>),
 
     // Public utility
@@ -268,7 +255,7 @@ struct DriverCallbacks<'ast> {
 }
 
 impl<'ast> DriverCallbacks<'ast> {
-    fn call_lint_level_at(&self, lint: &'static Lint, node: EmissionNode) -> Level {
+    fn call_lint_level_at(&self, lint: &'static Lint, node: EmissionNodeId) -> Level {
         (self.lint_level_at)(self.driver_context, lint, node)
     }
 
--- a/marker_api/src/context.rs
+++ b/marker_api/src/context.rs
@@ -105,24 +105,129 @@ impl<'ast> AstContext<'ast> {
         self.driver.call_lint_level_at(lint, node.emission_node_id())
     }
 
+    /// This function is used to emit a lint.
+    ///
+    /// Every lint emission, is bound to one specific node in the AST. This
+    /// node is used to check the lint level and as the main [`Span`] of the
+    /// diagnostic message. See [`EmissionNode`] for more information.
+    ///
+    /// The lint message, will be the main message at the start of the created
+    /// diagnostic. This message and all messages emitted as part of the created
+    /// diagnostic should start with a lower letter, according to [rustc's dev guide].
+    ///
+    /// The function will return a [`DiagnosticBuilder`] which can be used to decorate
+    /// the diagnostic message, with notes and help messages. The diagnostic message will
+    /// be emitted when the builder instance is dropped.
+    ///
+    /// [rustc's dev guide]: <https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-output-style-guide>
+    ///
+    /// ## Example 1
+    ///
+    /// ```
+    /// # use marker_api::prelude::*;
+    /// # marker_api::declare_lint!(
+    /// #     /// Dummy
+    /// #     LINT,
+    /// #     Warn,
+    /// # );
+    /// # fn value_provider<'ast>(cx: &AstContext<'ast>, node: ExprKind<'ast>) {
+    ///     cx.emit_lint(LINT, node, "<lint message>");
+    /// # }
+    /// ```
+    ///
+    /// The code above will roughly generate the following error message:
+    ///
+    /// ```text
+    ///  warning: <lint message>                <-- The message that is set by this function
+    ///  --> path/file.rs:1:1
+    ///   |
+    /// 1 | node
+    ///   | ^^^^
+    ///   |
+    /// ```
+    ///
+    /// ## Example 2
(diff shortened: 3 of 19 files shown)


marker_api/src/context.rs
```diff
@@ -112,30 +101,133 @@ impl<'ast> AstContext<'ast> {
...
+    ///   = help: <text>               <-- The added help message
+    /// ```
     #[allow(clippy::needless_pass_by_value)] // `&impl ToString`
-    pub fn emit_lint<F>(
+    pub fn emit_lint(
         &self,
         lint: &'static Lint,
-        node: impl Into<EmissionNode>,
+        node: impl EmissionNode<'ast>,
         msg: impl ToString,
```
xFrednet, line 216: I've tried a bit around with `impl ToOwned<Owned=String>` but couldn't get it to behave as nicely as `impl ToString` does. Is there a trick I've missed? :sweat_smile:

marker_api/src/diagnostic.rs
```diff
@@ -148,60 +219,167 @@ impl<'ast> DiagnosticBuilder<'ast> {
...
+/// determine the lint level and [`Span`] that is used for the main diagnostic
+/// message.
+///
+/// This trait is implemented for most AST nodes and node ids. When given the option,
+/// it's better to use the node directly, as the id might require some callbacks into
+/// the driver to fetch the actual node.
+//
+// FIXME(xFrednet): This trait should also be implemented for all ids that implement
+// `Into<LintEmissionId>`. However, for this we first need to add more methods to fetch
+// nodes by id, to provide a valid implementation for `emission_span`.
```
xFrednet, line 299: TODO @xFrednet Create an issue for this, once this PR is merged

marker_api/src/context.rs
```diff
@@ -112,30 +101,133 @@ impl<'ast> AstContext<'ast> {
...
+    ///     cx.emit_lint(LINT, node, "<lint message>").decorate(|diag| {
+    ///         // This closure is only called, if the lint is enabled. Here you
+    ///         // can create a beautiful help message.
+    ///         diag.help("<text>");
+    ///     });
+    /// # }
+    /// ```
+    ///
+    /// This will create the same help message as in example 2, but it will be faster
+    /// if the lint is enabled. The emitted message would look like this:
```
xFrednet, line 201:
```suggestion
    /// if the lint is allowed. The emitted message would look like this:
```

marker_api/src/context.rs
Veetaha: Use `impl Into<String>` instead. I'm not sure what you "nicety" you meant

marker_api/src/diagnostic.rs
```diff
@@ -13,24 +15,78 @@ use crate::{
...
+    /// From rustc a lint emission would look like this:
+    /// ```text
+    ///  warning: <lint message>
+    ///  --> path/file.rs:1:1
+    ///   |
+    /// 1 | node
+    ///   | ^^^^                 <-- The main span set by this function
+    ///   |
+    /// ```
+    pub fn set_main_span(&mut self, span: &Span<'ast>) -> &mut Self {
```
Veetaha, line 82: I wonder if this could have the same `impl EmissionNode` type that the constructor `emit_lint` function.

For example, [`darling::Error::with_span`](https://docs.rs/darling/latest/darling/struct.Error.html#method.with_span) uses a trait called `Spanned`, which is implemented for every `syn` AST node that has a `Span` and for `Span` itself. Could we use a similar approach? This would allow passing either the node, or the span

marker_api/src/diagnostic.rs
```diff
@@ -148,60 +219,167 @@ impl<'ast> DiagnosticBuilder<'ast> {
...
-            span: &self.span,
-            parts: parts.as_slice().into(),
-        };
-        cx.emit_diagnostic(&diag);
+        if self.emit_lint {
+            let parts: Vec<_> = self.parts.iter().map(DiagnosticPart::to_ffi_part).collect();
+            let span = self
+                .span
+                .as_ref()
+                .expect("always `Some`, if `DiagnosticBuilder::emit_lint` is true");
```
Veetaha, line 270: I'd make it this way:
```rust
struct DiagnosticBuilder {
    imp: Option<DiagnosticBuilderImp>
}

struct DiagnosticBuilderImp {
    // all fields are here
}
```
If the `imp` is `None` it means the lint shouldn't be emitted. It forces you at type level to write:
```rust
if let Some(diag) = &mut self.imp {
    // calculate the diagnostic here
}
```
When with `emit_lint` bool every field should have a trivially-constructible default value, and you also need to remember to put an `if self.emit_lint {}`.

marker_api/src/diagnostic.rs
```diff
@@ -148,60 +219,167 @@ impl<'ast> DiagnosticBuilder<'ast> {
...
+                    use $data_trait;
+                )*
+                self.id().into()
+            }
+
+            fn emission_span(&self, _cx: &$crate::AstContext<'ast>) -> Option<$crate::ast::Span<'ast>> {
+                $(
+                    use $data_trait;
+                )*
+                Some(self.span().clone())
```
Veetaha, line 332: Maybe if there were `Spanned { span() }` and `Identifiable { id() }` traits, then it would be possible to make this into a blanket impl instead of a macro boilerplate

Veetaha (commented):
Looks good

marker_api/src/context.rs
xFrednet: It's just my usual amount of typos, it was supposed to say nicely. I'll try this :D

marker_api/src/diagnostic.rs
xFrednet: That's a good idea :+1:

I have a question in this regard. Should the `Spanned::span` function replace the `span()` function of nodes, which already have one, or should it be an addition? Meaning:

```rs
struct Node {...}

impl Node {
    // Should this one be removed?
    // vvv
    fn span(&self) -> Span {...}
}

impl Spanned for Node {
	fn span(&self) -> Span {...}
}
```

Not having the function means that a user always has to import `Spanned` or `marker_api::prelude::*`. While having it, means that we have some code duplication. (Even if it's macro generated).

I'm always not 100% sure what's the best approach. I'm guessing, only providing the trait implementation, but some more input would be appreciated :)

marker_api/src/diagnostic.rs
xFrednet: The `Spanned` is a nice idea! I'm not sure about `Identifiable`, I've been thinking of adding a `NodeId` enum, which basically wrapps all possible node ids, then we would have `id()` for the node-specific id, like `ExprId` for expressions and `Identifiable::node_id() -> NodeId` which can wrap all IDs.

marker_api/src/context.rs
xFrednet: `Into<String>` seems to work well :D

marker_api/src/diagnostic.rs
Veetaha: +
I admit, that it's not clear to me either. I did it once on my experience where I created a trait, but also added the inherent impl to the type that just forwards the call to the trait impl. But in that case I only had one trait.

But here the question is whether it makes sense to have this in general for all traits that are already present in `marker` as well.

There are several degrees of following/not-following this approach.
1. We can make all traits private, and make all the traits methods inherent impl methods of the types. This significantly reduces the API surface of marker, meaning more freedom to `marker_api`'s code, but introduces boilerplate (even macro-generated one), and users lack generic traits for the purposes of their own utils.
2. We make all traits public, and don't provide the inherent impls. This hurts the user experience, because users need to import the traits. If there is a `prelude` module that imports all traits, it reduces the problem. In this case we expect users to always have a `use marker_api::prelude::*` at the top.
3. We make all traits public and provide the inherent impls. Users never need to use the `prelude`, but we are left with writing boilerplate inherent impls.
4. We add inherent impls only for some widely-used trait methods, meaning users *mostly* don't need to use `prelude`, but for some advanced stuff and generic programming - definitely.

I don't know what's best, because there isn't an obviously correct choice. Subjectively I'd lean to 2 and require users to always have a `prelude` import. 

<sub>I also found [this cool proc macro](https://github.com/dtolnay/inherent) from David Tolnay.</sub>

xFrednet: I think we should definitely provide public traits. So the decision is between 2, 3 and 4. I'm leaning towards 4. for really common traits, like `Spanned` would be.

---

I threw our conversation at ChatGBT and the response wasn't clear either. It basically listed the mentioned tradeoffs. It made one clear statement (It's ChatGBT, so take this with a grain of salt):
> Given your leaning towards Approach 2 (requiring trait imports), this approach seems reasonable, especially if it aligns with established practices in the Rust ecosystem.

Thinking about it, I can't think of any library that uses option 3. Probably in big part to avoid boilerplate and code duplication. If `#[inherent]` (from the linked crate) would be a compiler build in, it would likely be more common.

---

> We make all traits public and provide the inherent impls. Users never need to use the prelude, but we are left with writing boilerplate inherent impls.

The current codebase already implements several traits using `macro_rules!`. (The AST structure of Marker heavily relies on this.) Implementing the functions on the types themselves shouldn't be a big problem with this setup. So I believe, the boilerplate would be manageable, and if not, we could use a proc macro, like the linked one.

---

I've tried to google this question, but wasn't even sure how to phrase it correctly in "generic" terms. :sweat_smile:

For the `Spanned` trait, it feels like having both would be the best solution. But I'll leave the other traits and trait implementations as they are right now. If that also sounds reasonable to you?

marker_api/src/diagnostic.rs
xFrednet: I'm considering this solution or having everything lazy, how you proposed it here https://github.com/rust-marker/marker/pull/245#discussion_r1328026700

> Eagerly setting message for the allowed lint may also impact performance.
> I wonder if the API could then look like this which forces all users to calculate the diagnostics lazily.
> 
> ```rust
> // Simple case
> cx.emit_lint(LINT, node, |_| "Simple message");
> 
> // For elaborate case
> cx.emit_lint(LINT, node, |diag| diag.message("message").help("help"));
> ```
> 
> To implement this the closure needs to have a return type that implements some trait e.g. `Diagnostic`, where everything that implements `Into<String>` also implements `Diagnostic`, and the `DiagnosticBuilder` too.
> 
> But otherwise, the API in this form is also workable. I don't have a strong preference

I understood your comments in the feedback issue, that you found the closure rather confusing. In which case, the first option would be better. :thinking:

marker_api/src/diagnostic.rs
Veetaha: Yes, sounds reasonable :+1:

marker_api/src/diagnostic.rs
Veetaha: I don't see how putting everything under an `Option` in `DiagnosticBuilder` influences the `emit_lint()` method signature
So what signatures are you comparing against?

xFrednet: Well, if the closure taking the diagnostic builder is only executed, if the lint is emitted. We would only instantiate the diagnostic builder, to call the closure, meaning that the fields don't need to be conditional.

Veetaha: Nope, my suggestion was such that the `emit_lint` method would return `()`. I guess you are right, in case the closure is passed as a parameter to `emit_lint` then nothing optional is needed in the builder

xFrednet: Okay, the suggested API seems quite ergonomic. I'll probably think a bit more about this, before I move forward in either direction. Thank you for the input, I really appreciate it!

xFrednet:
I've now added the `Spanned` and `Identifiable` traits. I think they make sense, but they required me to touch numerous files. Oh well.

I had to rebase, due to some conflicts (Sorry). It probably makes sense to wait, with the review, until the next updated. The first commits stayed the same, anything after
"[Chore: Address PR Review Part 2](https://github.com/rust-marker/marker/pull/245/commits/061d06876734b908f89bc45fe4984135d613a370)" is new

marker_api/src/diagnostic.rs
xFrednet: I like having a trait for the last parameter of `emit_lint`, I've done some testing, but sadly got stuck. Do you maybe have any ideas on how to fix it? This is the trait and the implementations

```rs
pub trait IntoDiagnostic<'ast> {
    fn build_diagnostic(self, builder: DiagnosticBuilder<'ast>);
}

impl<'ast> IntoDiagnostic<'ast> for &str {
    fn build_diagnostic(self, mut builder: DiagnosticBuilder<'ast>) {
        builder.set_main_message(self);
    }
}
impl<'ast> IntoDiagnostic<'ast> for String {
    fn build_diagnostic(self, mut builder: DiagnosticBuilder<'ast>) {
        builder.set_main_message(self);
    }
}
impl<'ast, F: FnOnce(DiagnosticBuilder<'ast>)> IntoDiagnostic<'ast> for F {
    fn build_diagnostic(self, builder: DiagnosticBuilder<'ast>) {
        self(builder);
    }
}
```

With text it works just fine, but when I try to use a closure, I get the following error:

```
error[E0282]: type annotations needed
   --> marker_uilints/src/lib.rs:201:47
    |
201 |         cx.emit_lint(PRINT_EVERY_EXPR, expr, |mut diag| {
    |                                               ^^^^^^^^
202 |             diag.set_main_message("expr")
    |             ---- type must be known at this point
    |
help: consider giving this closure parameter an explicit type
    |
201 |         cx.emit_lint(PRINT_EVERY_EXPR, expr, |mut diag: /* Type */| {
    |
```

The type requirement makes sense to me, at the same time I would like this to be optional. Typing `: DiagnosticBuilder<'_>` seems less ergonomic. I assume this happens because `diag` is mutable. I've also tried having the type be `&mut DiagnosticBuilder<'_>` but that also requires the type annotation :thinking: 

Currently, I can think of three solutions:
1. Require the type annotation
2. Use internal mutability
3. I'm missing something :thinking: 

I pushed the last (broken) state with `&mut DiagnosticBuilder<'_>` as a parameter.

xFrednet: @Veetaha Do you maybe have an idea how to solve the problem?

I couldn't find a solution. Now, I'm thinking of having two methods:
1. `Context::emit_lint(<lint>, <node>, <message>)`
2. `Context::emit_lint_and_decorate(<lint>, <node>, <message>, <decoration closure>)`

Veetaha: Yeah, that's a tough problem. I know, for example, [reqwest_middleware::RequestInitialiser](https://docs.rs/reqwest-middleware/latest/reqwest_middleware/trait.RequestInitialiser.html) suffers from the same. This inference problem isn't specific to mutability, I tried `&` and by-value, but it still requires a type annotation.

I'll try to think more about this.

xFrednet: Thank you!

I've now implemented the second option `emit_lint` and `emit_lint_and_decorate` for comparison. It seems to work fairly well. It's a little less flexible but easier to use IMO. You can checkout the latest commit for that solution or [2204702](https://github.com/rust-marker/marker/pull/245/commits/2204702ec888d381439dbdab333a6d95d6d14639) for `IntoDiagnostic`

I'll fix the CI later, for now, it's just a showcase of the two options :)

xFrednet: > Maybe the diagnostic builder shouldn't expose the set_main_message method?

Yeah, I think that method should be removed. I wanted to reduce the number of arguments, but that one was a bit too aggressive.

> I was originally commenting that constructing the message could be non-trivial and should also be in closure, but maybe that's not actually a problem?

I liked the theoretical freedom of your proposed solution. The main messages themselves are usually very trivial. Only a tiny fraction of Clippy's have a dynamic main message, usually it's just a string literal.

> This will only matter when the lint is disabled, and there should generally be quite small amount of `allows` in code.

This highly depends on the lint authors and users of Marker. Currently, I only have Clippy as a reference, where most lints are allow-by-default. It can be expected most Marker lints will be warn-by-default. But if they have a higher false positive (FP) rate, they're also likely to be allowed more.  

> So I guess your original API is fine.

Okay, thank you for all the input! We can also adapt this later if needed.

> Regarding the difference between `emit_lint_and_decordate()` and `emit_lint().decordate()`. The latter looks cleaner but may be harder to discover.

We can always create a lint to suggest the usage of `.decorate()` :wink:

xFrednet:
This version now uses the `AstContext::emit_lint(lint, node, msg) -> DiagBuilder()` method.

I've used revert commits to not force push. Before the merge, I want to squash most commits to have a cleaner history. I'm generally happy with this rework, and especially about all the new function docs. :tada:

marker_api/src/ast/common/id.rs
```diff
@@ -144,3 +146,63 @@ new_id! {
...
+}
+
+impl_into_node_id_for!(Expr, ExprId);
+impl_into_node_id_for!(Item, ItemId);
+impl_into_node_id_for!(Stmt, StmtId);
+impl_into_node_id_for!(Body, BodyId);
+impl_into_node_id_for!(Field, FieldId);
+impl_into_node_id_for!(Variant, VariantId);
+
+pub trait Identifiable: Sealed {
```
Veetaha, line 185: I'm thinking of a convention for such single-method traits. For example, it's custom to name the trait exactly the same as the method that it wraps. Here are examples from std: `From { fn from() }`, `Clone { fn clone() }`, `Default { fn default() }`, `AsRef { fn as_ref() }`. But here are small inconsistencies: `IntoIterator { fn into_iter() }`.

But in this case, if we named the trait just `NodeId` then not only it wouldn't reflect the "verb" of what the trait does, it would also conflict with the `NodeId` enum because they share the same namespace. Maybe as an exception for "getter-like" traits they could be prefixed with `Get` like `trait GetNodeId { fn node_id() }`, and  `Spanned` could be named `trait GetSpan { fn span() }`. 

This is because `Identifiable` and `Spanned` sound rather ad-hoc. For example, if in the future we add a trait for `attributes()` method that would return a list of attributes for the node, then how would we call it `Attributesful` or `Attributed`? Well `Attributted` doesn't sound so bad, but maybe another example for an imaginary `children()` method would it be called `Childrenful`? Naming is hard.. 

Also `Identifiable` sounds like a rather overly generic trait name. What if in the future we add more traits like this, and the new ones will return a new type of ID? With `GetNodeId` such extensibility would be trivial, because the new traits will be called `Get{NewTypeOfIdHere}`.

marker_api/src/ast/common/span.rs
```diff
@@ -570,3 +570,31 @@ impl_ident_eq_for!(
     std::ffi::OsString,
     std::borrow::Cow<'_, str>
 );
+
+/// A trait for nodes, that provide a [`Span`].
```
Veetaha, line 574: What do you think about implementing this trait on `Span` type itself?

marker_api/src/diagnostic.rs
```diff
@@ -1,36 +1,74 @@
...
     lint: &'static Lint,
+    node: NodeId,
     msg: String,
-    node: EmissionNode,
     span: Span<'ast>,
     parts: Vec<DiagnosticPart<String, Span<'ast>>>,
 }
 
-#[allow(clippy::needless_pass_by_value)] // `&impl ToString` doesn't work
+#[allow(clippy::needless_pass_by_value)] // `&impl Into<String>` doesn't work
```
Veetaha, line 32: I'm not sure what you meant by this justification comment originally for `&imp ToString` doesn't work. What exactly doesn't work with it?

..but `&impl Into<String>` is definitely not going to work, because `Into<String>` consumes `self`, and when you have an `&impl Into<String>` you can't consume it, so if you ever happen to stumble with this type, then this type is totally useless, because you won't be able to call the only `into(self)` method available for it.

Anyway, I checked that if you remove this `allow()` clippy still passes. I suppose it's because there is no longer any `impl ToString` used in this impl block.

marker_api/src/diagnostic.rs
```diff
@@ -1,36 +1,74 @@
...
+    /// From rustc a lint emission would look like this:
+    /// ```text
+    ///  warning: <lint message>
+    ///  --> path/file.rs:1:1
+    ///   |
+    /// 1 | node
+    ///   | ^^^^                 <-- The main span set by this function
+    ///   |
+    /// ```
+    pub fn set_span(&mut self, span: &Span<'ast>) -> &mut Self {
```
Veetaha, line 64: Maybe this should be called just `span` (stripping the `set_` prefix)? Other methods like `note()` don't have `set_` prefix. 
For consistency it makes sense to either have either of the two:
- No `set_` prefix for setters. If getters are needed give them a `get_` prefix, like in `std::process::Command` API there is [`args()`](https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.args) setter and [`get_args()`](https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.get_args) getter.
- Add `set_` prefix for setters. If getters are needed give them no prefix.

The first approach looks better for builders where the bulk of the API usage is setters, and getters are almost never used or even non-existent.

Veetaha (approved):

marker_api/src/ast/common/id.rs
xFrednet: How about `Has<Ty>`, like `HasSpan` and `HasNodeId`? This makes it a property again. I think `Spanned` or `Identifyable` sounds a bit "better" when reading the code, but this won't really work for other things, like the attribute example you mentioned

I believe we took `Spanned` from the example of [`darling::Error::with_span`](https://docs.rs/darling/latest/darling/struct.Error.html#method.with_span).

marker_api/src/ast/common/span.rs
xFrednet: Yes, sounds good :)

marker_api/src/diagnostic.rs
xFrednet: I'm not sure if `span()` might be too generic and also too similar to the `span()` method from all other nodes. I would also argue that `.note()` is slightly different in the way that it adds a note and doesn't override an existing one :thinking:.

The builder example speaks again for just calling it `span()`. How strong do you feel about either option?

marker_api/src/ast/common/id.rs
Veetaha: Yeah, I remember rust-analyzer also had a bunch of traits `AttributesOwner` or `CommentsOnwer` etc. Then they renamed them to `HasAttributes` and `HasComments` etc. Idk, either of them works for me just like `GetAttributes`.
Anyway, I'll leave the final judgement call to you

marker_api/src/diagnostic.rs
Veetaha: Rather strong. I believe the builder should follow `std::process::Command` builder design. That type has, for example `.current_dir()` method that is a setter which if you call it twice, the second call will override the first one's value.

But the methods `.arg()`, `.args()` don't override the previous values, they just push new ones. And like this API, personally. It's quite obvious and intuitive for me. I'm glad that guys from `std` didn't add any extra prefixes/suffixes.

xFrednet: Okay, I don't have a strong opinion, so I'll take your suggestion :)

xFrednet:
This should hopefully be it. I'll squash the commits, before the merge :)

Veetaha (approved):

PR #246, opened 2023-09-17
Reorganize the CLI to a consistent module structure 

Veetaha:
Every CLI command now lives in a separate module. This helps with readability because code is now categorized via module files. Previously it was hard to read it because everything was in one file.

Replaced some leftover `OsString` handling with just `String`.

Replaced `HashMap` with `BTreeMap` in several places to have stable sorting. Meaning, for example, if you run `cargo marker` several times lint crates will be compiled in the same order (instead of a random one). See also this [doc](https://elastio.notion.site/Prefer-BTreeMap-Set-6042af3bc69041348c07170722b45bd1?pvs=4) for additional reasoning.

Changed some `pub` types to `pub(crate)` since this is a binary crate nothing should actually be `pub` here.

### DIFF
--- a/cargo-marker/src/backend.rs
+++ b/cargo-marker/src/backend.rs
@@ -9,12 +9,9 @@ use crate::config::LintDependencyEntry;
 use crate::error::prelude::*;
 use crate::observability::display::{self, print_stage};
 use crate::observability::prelude::*;
-use std::{
-    collections::HashMap,
-    ffi::{OsStr, OsString},
-};
-
 use camino::Utf8PathBuf;
+use itertools::Itertools;
+use std::collections::BTreeMap;
 
 pub mod cargo;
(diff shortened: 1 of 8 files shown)


xFrednet (approved):
Looks good to me, nice refactoring!

> Replaced HashMap with BTreeMap in several places to have stable sorting. Meaning, for example, if you run cargo marker several times lint crates will be compiled in the same order (instead of a random one). See also this [doc](https://elastio.notion.site/Prefer-BTreeMap-Set-6042af3bc69041348c07170722b45bd1?pvs=4) for additional reasoning.

An excellent reference, thank you! I believe that most cases in the rustc driver will benefit from hashmaps, as there is a huge number of nodes and requesting things by ID is the main operation, but that's something worth testing down the line.

PR #252, opened 2023-09-23
Add windows installation script

Veetaha:
ChatGPT rules.However of course it made a ton of mistakes converting bash to powershell, and I of course I had to stare at powershell syntax with their poor error messages for some time. But, anyway, it's past that and the script works.

You can test it using a separate branch where I configured it to work with my fork.

```ps1
curl.exe --location --silent --fail --show-error --retry 5 --retry-connrefused `
    https://raw.githubusercontent.com/Veetaha/marker/feat/install-on-windows-test/scripts/release/install.ps1 | powershell -command -
```

This was inspired by the installation script in [deno](https://github.com/denoland/deno). Turns out both `curl.exe` and `tar.exe` are [available on windows by default](https://learn.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows), and deno also uses them for its installation script.

There are some stupid caveats with using unicode symbols (I'm not even saying ansi colors) on windows. I didn't look into that, and just made the script monochrome and use `>` character for logging the executed commands.

---

I didn't update the docs here. I think I'll just make separate PR for that whicj will also include the commit with the docs for linux

---

Also fixed the behavior of `install.sh` when the `$CARGO_HOME` variable is set. Previously it was writing directly to it, but instead it has to write to a `$CARGO_HOME/bin` subdirectory.

### DIFF
--- a/scripts/release/install.ps1
+++ b/scripts/release/install.ps1
@@ -1,0 +1,165 @@
+# This script downloads the `cargo-marker` and the `marker_rustc_driver`
+# binaries from the GitHub release assets. It also sets up the required
+# Rust toolchain that the `marker_rustc_driver` depends on.
+#
+# This script must be self-contained! Nothing here should use anything from
+# the marker repository, because users are expected to run this script on
+# their machines where they don't have the marker repository cloned.
+#
+# It's possible to run this script on Linux if you have PowerShell installed there.
+# An important caveat is that you'll likely install the latest PowerShell 7.0 on linux,
+# but on real Windows 10/11 the version of PowerShell is quite old (5.1), so don't
+# try to use any new features.
+
+$ErrorActionPreference = "Stop"
+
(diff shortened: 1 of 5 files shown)


Veetaha:
**UPD**
I just made the script to work on linux too. It's possible to install powershell on linux. The only thing is that I had to add a condition that adds `.exe` to `curl` and `tar` commands on windows. So now it's possible to test the script on linux as well

Veetaha:
I'll just leave this here
https://github.com/PowerShell/PowerShell/issues/3223#issuecomment-1732439381

xFrednet (approved):
Looks good to me. I've tested it on a Windows computer, and it worked perfectly. Thank you for this!

Also, I have to say: I love the comments ^^ It makes reviewing the changes a lot more fun :heart:

xFrednet (approved):

xFrednet:
Thank you!

PR #255, opened 2023-09-25
Add docs about installing the precompiled binaries

Veetaha:
I decided to reduce the `curl` line with the short option names and removed `--retry 5 --retry-connrefused` because there are no short options for those. This makes the installation script a pure one-liner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,30 @@ The following is an abbreviated guide. Check out [The Marker Book] for detailed 
 
 ### Installation
 
+#### Download pre-compiled binaries (recommended)
+
+We provide pre-compiled binaries for the mainstream platforms. See the list of available artifacts in our [Github Releases](https://github.com/rust-marker/marker/releases/latest).
+
+Select one of the installation scripts below according to your platform. The script will install the required Rust toolchain dependency on your machine, download the current version of `cargo-marker` CLI, and the internal driver.
+
+<!-- region replace-version stable -->
+
+**Linux or MacOS (Bash)**:
+```bash
+curl -fsSL https://raw.githubusercontent.com/rust-marker/marker/v0.2.1/scripts/release/install.sh | bash
+```
+
+**Windows (PowerShell)**:
+```ps1
+curl.exe -fsSL https://raw.githubusercontent.com/rust-marker/marker/v0.2.1/scripts/release/install.ps1 | powershell -command -
+```
+
+<!-- endregion replace-version stable -->
+
+The provided scripts are pinned to a specific version of `marker` to avoid sudden breakages especially if this script will be used on CI.
+
+If you are a windows user or your platform isn't supported yet by the pre-compiled binaries, then you should fall back to building from sources as described below.
+
 #### Build from sources
 
 ```sh
--- a/scripts/release/snapshot.diff
+++ b/scripts/release/snapshot.diff
@@ -61,6 +61,14 @@
  # endregion replace-version dev
 
 === README.md ===
+ ```bash
+-curl -fsSL https://raw.githubusercontent.com/rust-marker/marker/v<version>/scripts/release/install.sh | bash
++curl -fsSL https://raw.githubusercontent.com/rust-marker/marker/v0.1.0/scripts/release/install.sh | bash
+ ```
+ ```ps1
+-curl.exe -fsSL https://raw.githubusercontent.com/rust-marker/marker/v<version>/scripts/release/install.ps1 | powershell -command -
++curl.exe -fsSL https://raw.githubusercontent.com/rust-marker/marker/v0.1.0/scripts/release/install.ps1 | powershell -command -
+ ```
  # An external crate from a registry
 -marker_lints = "<version>"
 +marker_lints = "0.1.0"
@@ -88,6 +96,16 @@
 -        cli_example = display::cli(r#"cargo marker --lints "marker_lints = '<version>'""#),
 +        cli_example = display::cli(r#"cargo marker --lints "marker_lints = '0.1.0'""#),
          lints = "--lints".blue(),
+
+=== docs/book/src/usage/installation.md ===
+ ```bash
+-curl -fsSL https://raw.githubusercontent.com/rust-marker/marker/v<version>/scripts/release/install.sh | bash
++curl -fsSL https://raw.githubusercontent.com/rust-marker/marker/v0.1.0/scripts/release/install.sh | bash
+ ```
+ ```ps1
+-curl.exe -fsSL https://raw.githubusercontent.com/rust-marker/marker/v<version>/scripts/release/install.ps1 | powershell -command -
++curl.exe -fsSL https://raw.githubusercontent.com/rust-marker/marker/v0.1.0/scripts/release/install.ps1 | powershell -command -
+ ```
 
 === docs/book/src/usage/lint-crate-declaration.md ===
  # An external crate from a registry
--- a/docs/book/src/usage/installation.md
+++ b/docs/book/src/usage/installation.md
@@ -11,6 +11,30 @@ The marker sub-command is provided by the *cargo_marker* crate. This crate requi
 [Cargo]: https://github.com/rust-lang/cargo/
 [rustup]: https://github.com/rust-lang/rustup/
 
+## Download pre-compiled binaries (recommended)
+
+We provide pre-compiled binaries for the mainstream platforms. See the list of available artifacts in our [Github Releases](https://github.com/rust-marker/marker/releases/latest).
+
+Select one of the installation scripts below according to your platform. The script will install the required Rust toolchain dependency on your machine, download the current version of `cargo-marker` CLI, and the internal driver.
+
+<!-- region replace-version stable -->
+
+**Linux or MacOS (Bash)**:
+```bash
+curl -fsSL https://raw.githubusercontent.com/rust-marker/marker/v0.2.1/scripts/release/install.sh | bash
+```
+
+**Windows (PowerShell)**:
+```ps1
+curl.exe -fsSL https://raw.githubusercontent.com/rust-marker/marker/v0.2.1/scripts/release/install.ps1 | powershell -command -
+```
+
+<!-- endregion replace-version stable -->
+
+The provided scripts are pinned to a specific version of `marker` to avoid sudden breakages especially if this script will be used on CI.
+
+If you are a windows user or your platform isn't supported yet by the pre-compiled binaries, then you should fall back to building from sources as described below.
+
 ## Build `cargo marker` plugin from sources
 
 Marker provides a new Cargo sub-command, that handles the driver installation, lint crate compilation, and checking process for you. To install it, simply use:


README.md
```diff
@@ -39,6 +39,30 @@ The following is an abbreviated guide. Check out [The Marker Book] for detailed
 
 ### Installation
 
+#### Download pre-compiled binaries (recommended)
```
xFrednet, line 42: This should also be duplicated to `cargo-marker`'s readme.

---

Once the next release is reached and I have a bit more time, I might ask CatGBT[^1] for a script that ensures that two regions in different files remain in sync 

[^1]: Okay I think that is my favorite misspelling of the month, and I'll stick with it ^^

README.md
```diff
@@ -39,6 +39,30 @@ The following is an abbreviated guide. Check out [The Marker Book] for detailed
...
+**Windows (PowerShell)**:
+```ps1
+curl.exe -fsSL https://raw.githubusercontent.com/rust-marker/marker/v0.2.1/scripts/release/install.ps1 | powershell -command -
+```
+
+<!-- endregion replace-version stable -->
+
+The provided scripts are pinned to a specific version of `marker` to avoid sudden breakages especially if this script will be used on CI.
+
+If you are a windows user or your platform isn't supported yet by the pre-compiled binaries, then you should fall back to building from sources as described below.
```
xFrednet, line 64: This still says that windows is not supported.

xFrednet (commented):
Two small nits, otherwise LGTM.

README.md
Veetaha: Ah, right, somehow missed to remove this paragraph

README.md
Veetaha: Hm, maybe `cargo-marker`'s readme could be a symlink to the root README? I think people actually don't read the docs on crates.io, they just go straight to the repo's docs (at least I do).

xFrednet: That might be a good option, why are close enough. The main README contains a bit more content, but it wouldn't be bad to have that in `cargo-marker`'s readme as well :)

Do you want to add the link in this PR?

xFrednet (approved):
LGTM, I'll keep this PR open until closely before the next release (Scheduled for 2023-10-05)

Veetaha:
I fixed a broken link in the changelog that I just noticed

README.md
Veetaha: I see that the `README.md` for `cargo-marker` details a bit more about what `cargo-marker` is, and I wouldn't like to lose all that information overwriting this file with a symlink to the root README. So I re-considered this and added a reference to the main repo readme file for details on the installation.

Did you mean adding a URL link as well, right?

xFrednet: I meant a file link, I didn't remember that the README had more details. In that case, a URL link should be sufficient. Now that we have precompiled binaries, it's likely that users will take that route and not install Marker from crates.io anyways :)

Veetaha:
I added docs for the Github Action to README, and a new chapter about CI to the mdbook.
I left the `cargo-marker/README.md`, but removed the duplication with the root `README.md` and left a reference to the repo instead.

docs/book/src/usage/ci.md
```diff
@@ -0,0 +1,125 @@
...
+### Git tags
+
+The git tag used for the GitHub Action selects the version of Marker that will be installed. There are several flavours of git tags:
+
+- Sliding `v0.2` *(recommended)*. Use this to get automatic patch updates.
+- Fixed `v0.2.1`. Use this if you want to pin a specific version if there is a regression in a new patch version. If there is one, please create a [new issue], patch versions must never break anything!
+
+<!-- endregion replace-version stable -->
+
+> ⚠️ Different `v0.{minor}` versions contain breaking changes. There is `v0` git tag, but because there are breaking changes on each minor release it doesn't make sense to use it. However once Marker reaches `1.0` minor versions will be compatible and `v1` tag will be stable.
```
xFrednet, line 26: Formulation suggestion:

```suggestion
> ⚠️ The minor versions before Marker `v1` contain breaking changes. While there is a sliding `v0` tag, it's highly recommended to include the minor version as well. This prevents uncontrolled CI breakage with every release
```

Side note: I plan to write a chapter on stability, once v0.3.0 has been released.

docs/book/src/usage/ci.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Inputs
+
+All inputs are optional, they only allow tweaking the default behavior.
+
+| Name         | Description                                               | Type      | Default |
+|--------------|-----------------------------------------------------------|-----------|---------|
+| `install-only` | Only install Marker and don't run `cargo marker` command. | `boolean` | `false` |
+
+### Example workflows
```
xFrednet, line 36: I think this chapter is missing a reference, which lints will be used. What are your thoughts on adding this paragraph here?

```
These example workflows will use the lint crates specified in the `Cargo.toml` file by default. Refer to the [*Lint Crate Declaration*](./lint-crate-declaration.md) section for more information.
```

Rendered:

> These example workflows will use the lint crates specified in the `Cargo.toml` file by default. Refer to the [*Lint Crate Declaration*](./lint-crate-declaration.md) section for more information.

docs/book/src/usage/ci.md
```diff
@@ -0,0 +1,125 @@
...
+
+#### Basic usage
+
+Checkout the repository code, install the toolchain, Marker, and start linting.
+
+```yml
+jobs:
+    rust-marker-lints:
+        runs-on: ubuntu-latest
```
xFrednet, line 45: Warnings will cause rustc and Marker to still exit with 0, for CI we would want to deny warnings:

```suggestion
        runs-on: ubuntu-latest
        env:
            RUSTFLAGS: --deny warnings
```

The `env` field could also be specified above the `jobs` section, like it's done in Marker's CI, but having it here should make copy-and-pasta :spaghetti: easier :)

Same for the example below.

docs/book/src/usage/ci.md
```diff
@@ -0,0 +1,125 @@
...
+        steps:
+            - uses: actions/checkout@v4
+            - uses: actions-rust-lang/setup-rust-toolchain@v1
+            - uses: rust-marker/marker@v0.2
+```
+
+#### Advanced usage
+
+If you need something more than just `cargo marker` you may use the action to only install Marker and run custom commands using `cargo marker` manually just like in your local dev environment.
```
xFrednet, line 54:
```suggestion
If you need something more than just the `cargo marker` command, you may use the action to only install Marker and then manually run the `cargo marker` command, just like in your local dev environment.
```

docs/book/src/usage/ci.md
```diff
@@ -0,0 +1,125 @@
...
+
+Here is an example of how you could limit the set of crates that you want to lint.
```
xFrednet, line 56:
```suggestion
Here is an example of how you could limit the set of crates that you want to lint. Refer to `cargo marker --help` for a full list of available options.
```

scripts/release/snapshot.diff
```diff
@@ -61,12 +61,26 @@
...
  ```
-
-=== cargo-marker/README.md ===
+ ```ps1
+-curl.exe -fsSL https://raw.githubusercontent.com/rust-marker/marker/v<version-major-minor>/scripts/release/install.ps1 | powershell -command -
++curl.exe -fsSL https://raw.githubusercontent.com/rust-marker/marker/v0.1/scripts/release/install.ps1 | powershell -command -
+ ```
+ 
+-The provided scripts use a sliding git tag `v<version-major-minor>`, to allow for automatic patch version updates, however a fixed tag `v<version-major>.2.1` is also available.
++The provided scripts use a sliding git tag `v0.1`, to allow for automatic patch version updates, however a fixed tag `v0.1.0` is also available.
```
xFrednet, line 74: The `v<version-major>.2.1` in this diff looks wrong. Shouldn't this be `v<version-major-minor>`? :thinking: 

The + line is correct again. This might just be a thing in the snapshot script.

scripts/release/snapshot.diff
```diff
@@ -89,6 +103,46 @@
...
+ ```
+ 
+-- Sliding `v<version-major-minor>` *(recommended)*. Use this to get automatic patch updates.
+-- Fixed `v<version>`. Use this if you want to pin a specific version if there is a regression in a new patch version. If there is one, please create a [new issue], patch versions must never break anything!
++- Sliding `v0.1` *(recommended)*. Use this to get automatic patch updates.
++- Fixed `v0.1.0`. Use this if you want to pin a specific version if there is a regression in a new patch version. If there is one, please create a [new issue], patch versions must never break anything!
+ 
+     --show-error \
+-    --retry <version-major> \
++    --retry 0 \
```
xFrednet, line 119: This could also become problematic. I wonder if it would be better to just ignore major versions and only do `major.minor` ones. :thinking: 

Or if we want to update the major one automatically, we could expect the `v` prefix? :thinking:

xFrednet (commented):
I have a few nits and formulation suggestions. Generally speaking, I like this update :D

scripts/release/snapshot.diff
Veetaha: Ah, right, good catch. I though I fixed all the edge cases, but seems not. I'll fix the script

cargo-marker/README.md
```diff
@@ -21,50 +21,11 @@
...
-```sh
-cargo marker
-```
+[Clippy]: https://github.com/rust-lang/rust-clippy
 
-This will initialize Marker, compile the lint crates and start linting.
+## Usage
 
-[Clippy]: https://github.com/rust-lang/rust-clippy
+See the installation and usage instructions in the [main Marker repository README][Marker].
```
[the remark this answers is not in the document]
Veetaha: Yeah, a link to the book is also fine

docs/book/src/usage/ci.md
Veetaha: Yeah, totally forgot about this one, thanks

scripts/release/snapshot.diff
Veetaha: Yeah, I've restricted major versions to `v{num}` syntax for now

Veetaha:
I've addressed the review comments and fixed the scripts.

I removed the `-E` option from `set -Eeuo pipefail`, because I noticed that with `-E` option the cleanup function may run twice or even three times depending on the nesting of subshell and shell functions.

I borrowed that idea from [this article](https://betterdev.blog/minimal-safe-bash-script-template/), which in turn borrowed it from [here](https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/).

It's notable how none of the articles mention the caveat that `set -E` will make the cleanup function run more than once if you have a subshell inside a back function like this:

```bash
#!/usr/bin/env bash
set -Eeuo pipefail

trap cleanup SIGINT SIGTERM ERR EXIT

NUM=0

function cleanup {
    trap - SIGINT SIGTERM ERR EXIT
    echo "Cleanup is running" >&2
}

function foo {
    $(non-existing-command)
}

foo

echo Finish
```

The output of this script is 
```
./dbg.sh: line 14: non-existing-command: command not found
Cleanup is running
Cleanup is running
```

If you remove the `-E` option, then the cleanup will run only once. The `trap EXIT` saves the day, because if the error happens in a subshel `trap EXIT` is still executed, and it is executed once for the script and not once for the subshell.

That's mindblowing, but that's how it works. TIL: never use `set -E`, use `trap EXIT` instead.

docs/book/src/usage/ci.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Self-hosted runners
+
+The action must work on self-hosted runners out of the box as well. We don't test all possible operating systems on our CI, but if your self-hosted runners use the OS version from the range of [OS images supported by managed GitHub Actions runners] or other compatible distributions, then it should work. If you have some exotic OS setup which breaks our GitHub Action we will be interested to hear about that in a [new issue].
+
+## Other CI systems
+
+If you don't use GitHub Actions CI, you can still benefit from the pre-made installation scripts that automate the downloading of the pre-compiled binaries on CI for you.
+
+These ones slightly differ from the scripts mentioned in the [installation chapter](installation.md#download-pre-compiled-binaries-recommended). They are more verbose for additional readability in the CI templates, and they also contain additional options to retry spurious network errors for stability on CI.
```
xFrednet, line 99: Super NIT: I think the "These ones" could be confusing, since we talked about installation scripts earlier.

```suggestion
These curl commands differ slightly from the scripts mentioned in the [installation chapter](installation.md#download-pre-compiled-binaries-recommended). They are more verbose for additional readability in the CI templates, and they also contain additional options to retry spurious network errors for stability on CI.
```

xFrednet (approved):

Issue #243, opened 2023-09-14
[Discussion] Rename `AstContext` -> `MarkerContext` or something similar?


xFrednet:
The [`AstContext`](https://docs.rs/marker_api/latest/marker_api/context/struct.AstContext.html) is a central struct in Marker's API and lint crates. This means that it contains several functions and will continue to grow. I'm now considering splitting it into multiple structs, to group functions better together. An example would be a struct which can fetch nodes from the driver.

When I considered this, I thought about a possible name for this struct and thought, that `AstContext` or `AstMap` would be ideal names for this specific one. However, the `AstContext` name is already the name of the main context. I think it would make sense to rename the main context struct to something else, which includes all driver functions, maybe one of these options?

1. `MarkerContext`
2. `DriverContext`
3. `LintingContext`

I have two questions in this regard:
1. Is such a breaking change still okay to do, at this point?
   * Right now I don't know of any public lint crates, that uses marker. This might be one of the last released where we can still "easily" do such a change
2. What name would fit the best?
   * I like `MarkerContext`, and it's unlikely that the name `MarkerContext` would be used somewhere else

PR #256, opened 2023-09-25
API: Rename `AstContext` -> `MarkerContext` (Let's break everything :boom:)

xFrednet:
This PR has two gigantic breaking changes:
* `AstContext` was renamed to `MarkerContext` (See rust-marker/marker#243)
* `Span` and related structs have been moved to the new `marker_api::span` module

If anyone uses Marker right now, they will definitely notice this change. But I believe this is worth it and still acceptable at this point.


CHANGELOG.md
```diff
@@ -23,17 +23,35 @@ The following components are considered to be internal and they are excluded fro
...
 - [#239]: GitHub releases now provide precompiled binaries of `cargo-marker` and `marker_rustc_driver`.
-- [#231]: Significantly improved error handling
+- [#252]: Marker now provides install scripts for linux, macos and windows
+
+### Breaking Changes
+- [#256]: Renamed `AstContext` -> `MarkerContext`
+- [#256]: Moved `marker_api::ast::common::span` -> `marker_api::span`
+- [#241]: Renamed `QuestionMarkExpr` -> `TryExpr`
+- [#244]: `StmtKind` and `PatKind` no longer wrap `Kind*` directly
+- [#245]: Takes less arguments and returns a `DiagnosticBuilder` instance
```
Veetaha, line 50:
```suggestion
- [#245]: `emit_lint()` takes less arguments and returns a `DiagnosticBuilder` instance
```

marker_api/src/lib.rs
```diff
@@ -32,12 +31,12 @@ pub use interface::{LintPassInfo, LintPassInfoBuilder};
 pub trait LintPass {
```
Veetaha, line 31: Since there is a separation between an early and late lint pass in the rustic driver, maybe it makes sense to also name this trait `LateLintPass` and then add an `EarlyLintPass` when support for pre-macro-expanded lints is added, or how do you think this will be handled?

Veetaha (approved):

marker_api/src/lib.rs
xFrednet: Yeah, macros are something which I've been dreading for a long time. They have two big problems:
1. The question of how they should be handled/represented
2. Rustc as very limited support for them. The mentioned pre expansion pass has been softly deprecated for quite some time. I also prefer the late lint pass by a long shot.

I've extracted this topic into https://github.com/rust-marker/design/issues/47 where I described how I think we could macros. None of this is set in stone, and I would really like to have more macro-specific use cases, before making any decision on this.

If Marker grows more, it's likely that we can make supporting changes in rustc. Clippy would also benefit from better macro support. But such a change would also require several design discussions with the compiler team.

---

I assume that you asked this question in the context of breaking changes, with the question, if we should rename this trait as well. The short answer is, that I would prefer having a single lint pass for Marker, which would mean that the name can stay as it is :)

xFrednet:
I've done a small update of Marker's readme, I hope this won't conflict with #255 

@Veetaha I've stumbled across the [`## Contributing`](https://github.com/rust-marker/marker#contributing) section again during this update. In this regard, I wanted to ask if you remember running into any trouble when you got started with contributing?

Veetaha:
The only trouble that I had was figuring out that I need to add this to the RA vscode config
```json
{
    "rust-analyzer.rustc.source": "discover"
}
```

xFrednet:
Nice catch! I've updated the documentation (What a productive way to procrastinate on some other things I should actually work on ^^)

PR #259, opened 2023-09-26
Add a github action

Veetaha:
Added a github action template that will allow anyone to include `marker` into their CI using the following code:

```yaml
- uses: rust-marker/marker@v0.3
```

This PR encompasses several important changes.

## New `action.yml` template

This is a dead simple composite github action template file that just runs a couple of bash lines. It is cross-platform, meaning the same syntax may be used for both `windows` and `unix`. The action handles the selection of the bash or powershell script internally.

It runs `cargo marker` command by default. That can be disabled with `install-only` option if there is some more elaborate use case than this. The action is pretty conservative in its API, as it exposes just this single input parameter for now. I believe in the future its API may grow. For example, it we may have `lints` input variable that would use toml-syntax to pass that as `--lints` input parameter. But I'm not sure that is actually needed. I believe that will be a really rare use case. 99% of the time simple `cargo marker` will suffice. We'll hear from anyone who uses the action if they want more builtin configurablity for the run command, otherwise if they are stuck they may use this syntax:
```yaml
- uses: rust-marker@v0.3
  with: { install-only: true }
- run: cargo marker --any-command-you-want
```

The action will install the version of marker specified in the git tag it was referenced with. To make the syntax higher work I had to amend the releases flow to support the sliding `v{major}` and `v{major}.{minor}` tags. 

## New sliding `v{major}` and `v{major}.{minor}` tags.

I updated the `release.md` documentation about this.

## A bunch of improvements and most importantly compatibility fixes in the installation scripts

I knew the installation scripts will break on Github Actions, because I didn't test them there. I only tested them on my home Ubuntu and Windows environments. So no wonder they broke when I ran them as part of the Github Action on CI on various OSes that Github Actions support today.

I also improved logging output of both scripts. They now have a consistent logging style for both bash and powershell. You can see example output on [CI in my fork](https://github.com/Veetaha/marker/actions/runs/6329481150/job/17189845125?pr=9) where I tested the Github Action.
image

## New CI tests for the Github Action

I extended the current CI with tests for the Github Action. With the new `github-action-test` CI job we test both the Github Action and the installation scripts.

Right now these jobs fail on `rust-marker` CI and they will be red until `0.3` is officially released and we have the pre-compiled binaries that these CI jobs actually depend on.
I would like to leave this job like this even if it is red for now. It won't block the merge of any PRs unless you add the new CI jobs as required in the repo settings, and this is what you should do after `0.3` is released.

The CI jobs run on every OS that Github Actions support today, and we'll need to maintain the list of `runs-on` platforms used in the job matrix manually. Unfortunately there isn't a syntax to expand the list of all OSes in GitHub, but anyway, this way it makes CI more stable, and it makes more explicitly that we need to switch our OSes support when a new one is added and we have to update the CI code.

## Documentation

No documentation was updated yet, mainly because the action isn't usable yet because we don't have pre-compiled binaries released.

However, I'll update the docs in https://github.com/rust-marker/marker/pull/255 a bit later this week

## Call to action

I'd like to ask you to announce the `0.3` release on Reddit only after you sanity check that the installation scripts and the Github Action work for this release manually. This is because I tested all of the things in my custom fork, and I had to do some manually replacements of the URLs and tags in my fork to make it work there temporarily and returned my changes back to `rust-marker` and `0.2.1` manually. Hopefully, I didn't make a mistake there, but a double-check won't hurt.

You may test that by running the following 4 versions of the installation scripts with the following URLs to also check that the sliding tags and master work:
```
https://raw.githubusercontent.com/rust-marker/marker/v0.3.0/scripts/release/install.sh
https://raw.githubusercontent.com/rust-marker/marker/v0.3/scripts/release/install.sh
https://raw.githubusercontent.com/rust-marker/marker/v0/scripts/release/install.sh
https://raw.githubusercontent.com/rust-marker/marker/master/scripts/release/install.sh
```
You may also install powershell if your don't have Windows, or if you have run powershell versions of these as well.

For the Github Action you may just create a draft PR in marker repo itself you may replace the `uses: ./` in the new CI job with the following combinations:
```
uses: rust-marker/marker@v0.3.1
uses: rust-marker/marker@v0.3
uses: rust-marker/marker@v0
```

The most important is `@v0.3` variant, because this ensures the updates of that tag will be compatible, and users will be recommended to use this syntax.

### Github Marketplace

I also figured out that you may publish the action to Github Marketplace. It's the website that Github uses to promote Github Actions. It often appears in search results in google when you search for some github action. It should be a good source incoming interest. To publish the action to marketplace you actually need to have a Github Release with that action. I'm not yet sure how exactly it works. Hopefully you'll have to do this once. So once you create a Github Release, you should go to edit that Github Release, and there you will see a checkmark at the top that suggests publishing the Github Action to the Marketplace like this:

image

> (actually if you do this for the first time it will ask you to accept some terms-and-conditions, but I suppose you'll figure this out).

I didn't try to publish the Github Action from my fork for testing, because I'm not sure if I can remove it from there and I wouldn't like to occupy the name of the action.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,0 +1,28 @@
+name: Rust Marker Linter
+description: GitHub Action to install and run the Marker linter for Rust 🦀
+branding:
+  icon: edit-3
+  color: white
+
+inputs:
+  install-only:
+    description: >
+      If set to `true` then the action will only install `cargo marker`,
+      and will skip running `cargo marker`. Use this if you want to run
+      something more complex than just `cargo marker`. If you think there
+      may be a frequent use case for running a different command then we will be
+      glad if you open a feature request issue for that to extend the action input
+      parameters.
+
+    default: 'false'
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - run: ${GITHUB_ACTION_PATH:?}/scripts/release/install.${{ runner.os == 'Windows' && 'ps1' || 'sh' }}
+      shell: bash
+
+    - run: cargo marker
+      if: ${{ inputs.install-only == 'false' }}
+      shell: bash
--- a/docs/internal/release.md
+++ b/docs/internal/release.md
@@ -34,7 +34,7 @@ The `sha256` sum is a small file that users may optionally download together wit
 This [`install.sh`](https://raw.githubusercontent.com/rust-marker/marker/v0.2.1/scripts/release/install.sh) script, can be used to automatically download and verify Marker's binaries.
 <!-- endregion replace-version stable -->
 
-## Operating system versions coverage
+### Operating system versions coverage
 
 We want to cover not only the mainstream operating systems, but also some reasonable range of their versions. For example, for `ubuntu` we want our binaries to work on the current LTS version of `ubuntu` and also on the version of `ubuntu` that precedes it. At the time of this writing the latest LTS version of `ubuntu` is `22.04`, and the previous one is `20.04`, while the LTS version of ubuntu before that one is `18.04` and it already reached the "end of standard support" date, and so far practically noone uses it.
 
@@ -62,6 +62,21 @@ The binaries with `musl` in their name don't depend on GLIBC. They are staticall
 
 This comes at a cost, though. The `musl` implementation of `libc` isn't complete, and it may have bugs, performance degradations compared to GLIBC, etc. However, it generally covers everything you may ever need if you aren't doing something unusual.
 
+## Github Action
+
+There is a Github Action template `action.yml` file maintained within the Marker repository. It installs the pre-compiled binaries using the installation scripts and runs `cargo marker`.
+
+It is common to reference the Github Action using sliding tags like this:
+```yml
+- uses: rust-marker/marker@v0.3
+```
+Or this when marker reaches a `1.0.0` version and minor versions are compatible:
+```yml
+- uses: rust-marker/marker@v1
+```
+
+We maintain these tags in our automated release flow described below.
+
 # Regular release
 
 The regular release means a planned event when the maintainer of `marker` publishes the new version of the artifacts to the users. It may happen at any time when such a decision is made, which usually means on some consistent schedule.
@@ -80,7 +95,17 @@ Once the `CHANGELOG.md` is ready in `master` the maintainer can trigger the CI `
 
 The `release` CI workflow then checks out the `master` branch (assuming "Use workflow from" input wasn't changed from its default), parses the current release version from the `CHANGELOG.md` file and updates `Cargo.toml`, `Cargo.lock` and various `.rs` and `.md` files with the new version. It uses simple regex patterns with `sed` to edit the files. This logic may break, and thus there is a test on regular CI that makes sure it stays stable.
 
-The release commit is then assigned a `v{semver}` git tag. After that the workflow sets the next version for the new development cycle with the incremented release version and the `-dev` suffix, and creates a new commit with that. No additional tag at this point is created.
+The release commit is then assigned a `v{semver}` git tag . After that the workflow sets the next version for the new development cycle with the incremented release version and the `-dev` suffix, and creates a new commit with that.
+
+### Sliding `v{major}` and `v{major}.{minor}` tags.
+
+The release CI flow will also create and move the sliding `v{major}` and `v{major}.{minor}` tags. For example, if we release a version `0.3.0`, then there will be three tags created `v0.3.0`, `v0` and `v0.3`.
+
+And if we release a patch `0.3.1`, then there will be a new tag `v0.3.1`, and the existing `v0` and `v0.3` will be moved to this new  commit for `0.3.1`.
+
+The sliding tags won't be updated for pre-releases (versions with `-` suffix in them).
+
+These sliding tags may be used to hardcode only the major or minor version of `marker`, to allow for automatic updates, to the patch version. These sliding tags can be used to download the installation script or run Marker's Github Action and automatically get fixes from the upstream.
 
 ## `release-on-tag` workflow
 
--- a/scripts/release/install.ps1
+++ b/scripts/release/install.ps1
@@ -1,3 +1,5 @@
+#!/usr/bin/env pwsh
+#
 # This script downloads the `cargo-marker` and the `marker_rustc_driver`
 # binaries from the GitHub release assets. It also sets up the required
 # Rust toolchain that the `marker_rustc_driver` depends on.
@@ -14,6 +16,30 @@
 # This script is specifically for windows, but has a similar structure to the
 # unix `install.sh` script. If you modify this script, please check if the modifications
 # should also apply to the unix one.
+#
+# General PowerShell notes:
+#
+# Since we have $ErrorActionPreference = "Stop" the script will stop at any `Write-Error`.
+# Yeah, writing an error log counts as an error. Who could expect that? Regardless of how
+# unintuitive it is we use `Write-Error` instead of `throw` because the error that is generated
+# this way is more readable, because it refers to the call site of the function where the error
+# was written and not to the `throw` statement itself in the code snippet that is output.
+#
+# Example error output with `throw` shows the `throw` statement itself, not the call site:
+# ```
+# Line |
+#   74 |          throw "Command $cmd failed with exit code $LASTEXITCODE"
+#      |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#      | Command curl failed with exit code 22
+# ```
+# Example error output with `Write-Error` shows the call site of the function which is more useful:
+# ```
+# Line |
+#  174 |      Step curl$exe `
+#      |      ~~~~~~~~~~~~~~~
+#      | Command curl failed with exit code 22
+# ```
+
 
 $ErrorActionPreference = "Stop"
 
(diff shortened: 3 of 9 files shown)


action.yml
```diff
@@ -0,0 +1,28 @@
+name: Rust Marker Linter
+description: GitHub Action to install and run Marker linting for Rust 🦀
+branding:
+  icon: edit-3
+  color: white
```
Veetaha, line 5: Branding color and icon are used for GitHub Marketplace. You may change as you wish according to [the docs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#brandingcolor) there is a selection of colours and Icons that you can preview

Veetaha:
I mistakenly created a PR in this repo yesterday, when it was not ready for review.
Now it's ready for review

xFrednet:
> For example, it we may have lints input variable that would use toml-syntax to pass that as --lints input parameter. But I'm not sure that is actually needed. I believe that will be a really rare use case. 99% of the time simple cargo marker will suffice.

I like the idea of exposing the `lint` parameter, but agree that this will probably not be used anytime soon. Right now, it's the technical idea and elegance of the idea, that I like. The current setup sounds excellent to me :)

> Unfortunately there isn't a syntax to expand the list of all OSes in GitHub, but anyway, this way it makes CI more stable, and it makes more explicitly that we need to switch our OSes support when a new one is added and we have to update the CI code.

Do you know if there is a repo or something we can subscribe to, to get notified about relevant changes like a new OS being added? (Without our inbox being flooded with unrelated notifications.)

> No documentation was updated yet

The screenshot shows that it discovered a readme file, related to the action template. I'm assuming that it find the main readme of the repo. Do you know if it's easily possible to use a different file, which focuses on the GH action instead?

A related question: Numerous other projects have the action template in a separate repo instead of their main one. Have you considered the tradeoffs between having it in this repo or creating a separate one? Updating the version and tag is obviously a lot easier, since this repo already has the infrastructure for it, but I wonder if there are other things :thinking: 

> I'd like to ask you to announce the 0.3 release on Reddit only after you sanity check that the installation scripts and the Github Action work for this release manually. This is because I tested all of the things in my custom fork, and I had to do some manually replacements of the URLs and tags in my fork to make it work there temporarily and returned my changes back to rust-marker and 0.2.1 manually. Hopefully, I didn't make a mistake there, but a double-check won't hurt.

Yes! Something like this has been on my todo list. I plan to migrate our lint crate template as well, which will act like a second sanity check. That crate has already help to catch so many bugs. :bug: 

For the announcement: I still plan to do the announcement for v0.3.0 on Reddit to reach a larger audience. But I intended to hold off until a few days after the release. At first, this wasn't my plan, but I believe this to be the better approach than to rush the release, template migration and testing. People will also focus on the new Rust release first. My current dream schedule is as follows:
1. Push the last breaking changes and CI updates into the finish line
2. Publish v0.3.0 on 2023-10-05
3. Test everything and migrate the template
4. Add more documentation if there is time
5. Announce Marker on Tuesday 2023-10-10

I think having those five days will be a good thing, while not putting it too far into the future. 

> I also figured out that you may publish the action to Github Marketplace.

That sounds awesome. As usual, thank you for the excellent documentation!

---

I'll try to review this soon, but it might need to wait until the weekend. RL is currently taking up sometime, and I'm also trying to open the last big breaking changes PRs. (I'm debating of moving `marker_utils` into `marker_api` after the discussion we had in the other PR, but that's probably a discussion for a different PR :))

A sincere thank you for this PR, you have basically single-handedly reworked Marker's CI and important infrastructure in the last couple of weeks. :heart:

Veetaha:
> Do you know if there is a repo or something we can subscribe to, to get notified about relevant changes like a new OS being added? (Without our inbox being flooded with unrelated notifications.)

Yeah, there is [this repo](https://github.com/actions/runner-images#image-releases), but seems to have quite often releases with just pure tools version bumps, which is going to be quite chatty. GitHub also has twitter and own blog, but they also post ton of assorted posts there almost every day. So I would just be unnotified about this. I don't think it's very important for the marker Github Action tests to stay very up-to-date with the new Github Actions OS images. New images of major OS versions are released very rarely. Also, for example, when ubuntu 24.04 comes out I'm sure you'll know about that, and you'll know to expect GitHub to pick that up soon for the runners. So this comes down to just lazily checking the availability of now OS images every half a year or less. 

> Do you know if it's easily possible to use a different file, which focuses on the GH action instead?

Unfortunately, not, the README for the marketplace is not configurable.

> Have you considered the tradeoffs between having it in this repo or creating a separate one? 

Yes, I did. I decided to keep the action in the same repo. I know [GitHub recommends](https://docs.github.com/en/actions/creating-actions/about-custom-actions#choosing-a-location-for-your-action) using a separate repo for the ability of versioning and releasing the action separately. But having the action in the repo is easier at least at this stage. Also, given how simple the action actually is, I think it's fine to couple its release with the release of other marker code. 

I know of at least of [crate-ci/typos](https://github.com/crate-ci/typos) hosting the Github Action right within the CLI tool repo, same for [cargo-machete](https://github.com/bnjbvr/cargo-machete).

On the other hand [cargo-deny-action](https://github.com/EmbarkStudios/cargo-deny-action) is separate from the main `cargo-deny` repo, and I find it confusing for me, because the action has versioning different from `cargo-deny` itself. Also having a separate repo means you can't do code changes to the action and the installation scripts at the same time, so it brings more complexity in synchronizing changes. I generally like the synchronous versioning where `cargo-marker`, rust crates, the github action all have the same version. It makes it easier to reason about for users, although couples the release of them (so you can't make a patch to the action without releasing `cargo-marker` and crates). Let's see how it works and if there will be pain in this regard it's possible to just create a new repo for the action and ask everyone to update the action references.

I may not see the whole picture because I didn't develop a GitHub Action before, but for this simple action this should suffice.

action.yml
```diff
@@ -0,0 +1,28 @@
+name: Rust Marker Linter
+description: GitHub Action to install and run Marker linting for Rust 🦀
```
xFrednet, line 2:
```suggestion
description: GitHub Action to install and run the Marker linter for Rust 🦀
```

action.yml
```diff
@@ -0,0 +1,28 @@
...
+      glad if you open a feature request issue for that to extend the action input
+      parameters.
+
+    default: 'false'
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - run: ${GITHUB_ACTION_PATH:?}/scripts/release/install.${{ runner.os == 'Windows' && 'ps1' || 'sh' }}
```
xFrednet, line 23: Since this doesn't need to download the script, I wonder if it "clones" the entire repo for perform the composition. That would explain the advice, to have the action separate from the main repository :thinking: 

The Marker repo is currently about 7mb, which is reasonable, and would be fine for now. But it would be a good thing to know. Do you have any information about this? Or can we maybe test this, by simply running "ls"?

docs/internal/release.md
```diff
@@ -80,7 +95,17 @@ Once the `CHANGELOG.md` is ready in `master` the maintainer can trigger the CI `
...
+
+### Sliding `v{major}` and `v{major}.{minor}` tags.
+
+The release CI flow will also create and move the sliding `v{major}` and `v{major}.{minor}` tags. For example, if we release a version `0.3.0`, then there will be three tags created `v0.3.0`, `v0` and `v0.3`.
+
+And if we release a patch `0.3.1`, then there will be a new tag `v0.3.1`, and the existing `v0` and `v0.3` will be moved to this new  commit for `0.3.1`.
+
+The sliding tags won't be updated for pre-releases (versions with `-` suffix in them).
+
+These sliding tags may be used to not hardcode the version of `marker`, and allow for automatic patch updates, for example. They will most likely be referenced in the installation scripts download command and in Github Actions.
```
xFrednet, line 108: This one confused me a bit. This formulation is clear to me. I hope it contains the same information:

```suggestion
These sliding tags may be used to only hardcode the major or minor version of `marker`, to allows automatic updates, ot the patch version. These sliding tags can be used to download the installation script or run Marker's Github Action.
```

scripts/release/install.ps1
```diff
@@ -14,6 +16,30 @@
...
+#   74 |          throw "Command $cmd failed with exit code $LASTEXITCODE"
+#      |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#      | Command curl failed with exit code 22
+# ```
+# Example error output with `Write-Error` shows the call site of the function which is more useful:
+# ```
+# Line |
+#  174 |      Step curl$exe `
+#  |      ~~~~~~~~~~~~~~~
+#  | Command curl failed with exit code 22
```
xFrednet, line 40: NIT:

```suggestion
#      |      ~~~~~~~~~~~~~~~
#      | Command curl failed with exit code 22
```

scripts/release/install.ps1
```diff
@@ -68,7 +128,8 @@ function Check-Sha256Sum {
     $expected = Get-Content "$file_stem.sha256"
 
     foreach ($line in $expected) {
-        $line -match '(\S+)\s*\*?(.*)'
+        # Silence the output from the match, otherwise it prints `True` or `False`
+        $null = $line -match '(\S+)\s*\*?(.*)'
```
xFrednet, line 132: PowerShell amazes me, similar to how a scientist would be interested in a yet unknown animal species. I don't want to really roast PowerShell, but it just sounds so chaotic.

xFrednet (commented):
This version looks good to me, I have a few notes and still want to test the PowerShell script, before merging this. I should be able to test it sometime this weekend :)

---

I have one comment on the `install.sh` script, which is unrelated to the changes made in this PR. I wanted to test the modified script and ran `sh ./install.sh` and got an error. This [StackOverflow](https://stackoverflow.com/questions/5725296/difference-between-sh-and-bash) explains that `sh != bash`. Now, I wonder if the file should have a `.bash` ending instead of `.sh`. I know that unix doesn't care too much about file endings, but as a user I would first run sh instead of bash. :thinking: 

For reference, this is the error I got:
```
$ sh ./install.sh 
./install.sh: 15: set: Illegal option -o pipefail
```
While `$ bash ./install.sh` works perfectly fine :)

---

Either way, thank you for writing this action!

action.yml
xFrednet: I like the icon and color you picked out, so I'm in favor of keeping these :D

action.yml
Veetaha: I suppose it does a [sparse-checkout](https://git-scm.com/docs/git-sparse-checkout) or smth similar. I'll check what the file tree there is, and if there even is any `.git` dir there. Even if it's 7MB, I think it's nothing compared to downloading the pinned nightly rust toolchain (although with `--profile minimal`), which when decompressed is `544MB`. 
But that's something to keep in mind, definitely

scripts/release/install.ps1
Veetaha: Yeah, there is so much weird stuff here, but I'm sure everything has its explanation and even good reasoning that no longer makes sense in the modern era, but could be at hype years ago. E.g. I suppose powershell was born from the OOP hype with all the structured data, try/catch and dot net introp. That sounds cool, but also useless for the kind of the tasks it is used to day, which are probably "install program X, and run it, and be as thin as possible, because program X is actually cross-platform while you aren't :-1: "

Veetaha:
I've checked the file tree under the action path in a test [CI run](https://github.com/Veetaha/marker/actions/runs/6364567765/job/17280922635#step:3:11). There is no `.git` repository there, but there are all files for the repo there, so this isn't a spare checkout. I suppose github could use some internal API to download the repo without the git cruft.

The resulting repo size is 2.6 MB (see output from `du` [below](https://github.com/Veetaha/marker/actions/runs/6364567765/job/17280922635#step:3:79)). The download happens in the [job setup stage](https://github.com/Veetaha/marker/actions/runs/6364567765/job/17280922635#step:1:26), but there are no details of how that is implemented, but since it takes under a second that's not a problem today.

xFrednet (approved):
This version looks good to me! I tried it out, and everything works like a charm :tada: 

I would like to hold off on merging this, since it will make the CI and master red. I'll probably merge it on Tuesday or Wednesday if that's alright with you?

You can already update the documentation in https://github.com/rust-marker/marker/pull/255. I plan to merge that PR shortly before the release.

Veetaha:
Okay, I guess it's fine. Hopefully there won't be merge conflicts, but if there are any I'll try to fix them. If I don't respond quickly enough you have write access to my fork according to this setting on the PR (which is by default though):
image


I already updated the docs in https://github.com/rust-marker/marker/pull/255, you can take a look there.

xFrednet:
Now that I kicked off the merge of #262 I think we can merge this as well. It won't make a difference, and that's one less thing to keep track of. Sorry for the back and forth. Would you mind rebasing one last time?

I have one last big PR planned before v0.3.0 that will move all [semantic structs](https://docs.rs/marker_api/latest/marker_api/?search=Sem) from the `marker_api::ast` to a new `marker_api::semantic` module. The CI won't matter for that PR :)

Btw. Do I remember correctly, that the release CI pushes the commits directly to master? In that case, I have to check the settings that the branch protection won't prevent that :)

Veetaha:
Yes, you are right, you need to make sure that maker CI account can bypass the branch protection rules for `master`. There is definitely an option to make exceptions for particular users.

Veetaha:
I rebased an squashed everything into a single commit.

xFrednet (approved):

xFrednet:
Perfect, thank you very much :heart:

xFrednet:
> Yes, you are right, you need to make sure that maker CI account can bypass the branch protection rules for master. There is definitely an option to make exceptions for particular users.

image

I found this. The required status badge thing confuses me a bit. Maybe it buffers the commit and runs the CI on it? IDK, I'll read up on it until Thursday and maybe do an early check by adding @xFrednet and pushing something with that setting. :)

Veetaha:
I think this is the right setting according to the settings in other repos that I have. It should work

PR #260, opened 2023-09-27
API: Add `AstMap` (`MarkerContext` cleanup and more breakage :bomb:)

xFrednet:
This PR cleans up a central part of Marker, the callbacks to the driver. With the old `AstContext` continuously growing, I had the feeling that it was getting more and more overwhelming to work with. It only had a few public functions, but they had a whole range of functions. The old `AstContext` was basically a box that every tool was thrown into, without much consideration. It was like storing the stone drill next to the sewing kit and toaster. Definitely possible, but probably not useful.

A previous PR already renamed `AstContext` to `MarkerContext` for reasons described in rust-marker/marker#243.

This PR tried to add new structure to the while `context` module. From now on, we'll try to group similar functions together. Some general functions, like `emit_lint` will be available methods of `MarkerContext` directly. But other methods, requesting notes or checking something AST related, are available in a separate `AstMap` instance. This one can be accessed via the `MarkerContext::ast` method.

This separation is not perfect, and most internal callbacks are still stored inside `MarkerContext` itself. They should be moved at some point, but this PR focussed on the user-facing changes.

---

I've also added some documentation about the whole callback infrastructure. For me, the whole thing is like a zombie :zombie:. The existence both amazes and terrifies me.

Anyways, for anyone reading this, I hope you have a fantastic day, full of green CIs and friends to talk to :D

---

Closes: rust-marker/marker#243

### DIFF
--- a/marker_adapter/src/context.rs
+++ b/marker_adapter/src/context.rs
@@ -1,12 +1,101 @@
+//! The [`marker_api`] crate is designed to be driver-independent and lightweight.
+//! Communication from the lint-crate to the driver is done via structs inside the
+//! [`marker_api::context`] module. For example, [`marker_api::MarkerContext`].
+//!
+//! The communication needs to be ABI safe, since it goes over an FFI function
+//! boundary. For this reason, it's not (directly) possible to use `dyn Trait`objects.
+//! Marker instead uses `*Callback` structs, that contain function pointers, which can
+//! be filled by the driver. This is like a `dyn Trait` object, without all the nice
+//! syntax and build in compiler support.
+//!
+//! These `*Callbacks` structs need to be filled and some ABI-specific transformations
+//! have to be performed. This module hides all the gore behind some structs and
+//! traits, which can be implemented by the specific driver. The following is an
+//! explanation of this structure and some naming conventions. The explanation will use
+//! an imaginary `Magic` struct, to illustrate the structure:
+//!
+//! * The `marker_api` crate defines the `Magic` struct with an interface for external and internal
+//!   consumers.
+//!
+//!    It also defines the `MagicCallbacks` struct, which stores function pointers.
+//!    All `*Callbacks` structs have a special `data: &'ast MarkerContextData` field that will be
+//!    passed to all stored function pointers, as the first argument. All function
+//!    pointers are marked as `extern "C"` and need to use FFI safe types. Lifetimes
+//!    are allowed by Rust, but not enforced over FFI bounds.
+//!
+//! * The `marker_adapter` crate defines a `MagicDriver` trait, that has all the functions, needed
+//!   to provide a backend for the `Magic` struct.
+//!
+//!    It also defines the `MagicWrapper` struct, that has a `driver: &'ast dyn
+//!    MagicDriver` field. This struct will be used to fill the `MagicCallbacks.data`
+//!    field. Wrapping it in a separate `*Wrapper` struct makes the `&'ast *Wrapper`
+//!    pointer a thin pointer and cleans up the interface.
+//!
+//!    The `marker_adapter` module containing the `MagicWrapper` struct defines a
+//!    bunch of `extern "C"` functions, which build the counterpart to the `extern
+//!    "C"` functions inside `MagicCallbacks`. These functions cast the `data`
+//!    argument into the `&'ast MagicWrapper` instance it originated from and calls
+//!    the corresponding trait function from the `driver: &'ast dyn MagicDriver` field
+//!    stored inside the `Wrapper` struct. The `extern "C"` functions are also
+//!    responsible for converting all types into FFI safe types and back.
+//!
+//! * The driver simply implements the `MagicDriver` trait and instantiates `MagicWrapper`. This
+//!   wrapper instance is then used to fill `MagicCallbacks` and instantiate the `Magic` struct from
+//!   `marker_api`.
+//!
+//! ---
+//!
+//! Short Q&A with @xFrednet
+//!
+//! * **Isn't there a simpler way?**
+//!
+//!     Most likely... Or better say hopefully there is.
+//!
+//!     However, I couldn't find one. Most libraries I found either didn't support these
+//!     types of callbacks, were experimental, or are unsound.
+//!
+//!     I also have high hopes for Rust adding some cross crate boundary communication
+//!     support some time in the future. But there is no point in waiting on this. There
+//!     are lints which need to be written, and we can always replace this infrastructure
+//!     later.
+//!
+//! * **Isn't there a simpler way to make the types FFI safe?**
+//!
+//!     Not really. There are some tools to generate ABI safe types, but they are geared
+//!     towards C/C++ consumers. As a result, they usually lose lifetime information,
+//!     which we want to keep since both sides are using Rust.
+//!
+//!     There are also crates that rely on serialization. A nice and simple solution, but
+//!     not feasible for an entire AST over multiple FFI boundaries for every lint crate.
+//!
+//! * **Do you use code generation for all this infrastructure?**
+//!
+//!     I would love to! However, I haven't found the time to implement this. Normal macro
+//!     rules are basically out of the question due to all the required FFI type
+//!     transformation.
+//!
+//!     If anyone is interested in implementing this, I would be grateful!!! See
+//!     rust-marker/marker#122
+//!
+//! * **Is this implementation even safe and sound?**
+//!
+//!     Theoretically speaking? From my understanding? Yes, it is, assuming that both sides
+//!     reconstruct the lifetimes correctly.
+//!
+//!     Practically speaking? It wouldn't surprise me if there were several bugs. So far,
+//!     it has been working suspiciously well, but I won't complain.
+
 // The lifetimes are destroyed by unsafe, but help with readability
 #![allow(clippy::needless_lifetimes)]
 
+mod map;
+pub use map::*;
+
 use marker_api::{
     ast::{ty::SemTyKind, ExpnId, ExprId, SpanId, SymbolId},
-    context::DriverCallbacks,
+    context::{MarkerContextCallbacks, MarkerContextData},
     diagnostic::Diagnostic,
     ffi::{self, FfiOption},
-    lint::{Level, Lint},
     prelude::*,
     span::{ExpnInfo, FileInfo, FilePos, SpanPos, SpanSource},
 };
@@ -17,30 +106,26 @@ use marker_api::{
 /// change and calling functions on them would require a stable ABI which Rust
 /// doesn't provide.
 ///
-/// In this case, the [`DriverContextWrapper`] will be passed as a `*const ()`
-/// pointer to [`DriverCallbacks`] which will do nothing with this data other
+/// In this case, the [`MarkerContextWrapper`] will be passed as a `*const ()`
+/// pointer to [`MarkerContextCallbacks`] which will do nothing with this data other
 /// than giving it back to functions declared in this module. Since the `&dyn`
 /// object is created, only used here and everything is compiled during the same
 /// compiler run, it should be safe to use `&dyn`.
 #[repr(C)]
-pub struct DriverContextWrapper<'ast> {
-    driver_cx: &'ast dyn DriverContext<'ast>,
+pub struct MarkerContextWrapper<'ast> {
+    driver: &'ast dyn MarkerContextDriver<'ast>,
 }
 
-impl<'ast> DriverContextWrapper<'ast> {
-    #[must_use]
-    pub fn new(driver_cx: &'ast dyn DriverContext<'ast>) -> Self {
-        Self { driver_cx }
+impl<'ast> MarkerContextWrapper<'ast> {
+    pub fn new(driver: &'ast dyn MarkerContextDriver<'ast>) -> Self {
+        Self { driver }
     }
 
     #[must_use]
-    pub fn create_driver_callback(&'ast self) -> DriverCallbacks<'ast> {
-        DriverCallbacks {
-            driver_context: unsafe { &*(self as *const DriverContextWrapper).cast::<()>() },
-            lint_level_at,
+    pub fn create_callbacks(&'ast self) -> MarkerContextCallbacks<'ast> {
+        MarkerContextCallbacks {
+            data: unsafe { &*(self as *const MarkerContextWrapper).cast::<MarkerContextData>() },
             emit_diag,
-            item,
-            body,
             resolve_ty_ids,
             expr_ty,
             span,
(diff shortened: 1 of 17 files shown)


marker_adapter/src/context.rs
```diff
@@ -54,92 +135,72 @@ impl<'ast> DriverContextWrapper<'ast> {
...
 }
 
 extern "C" fn symbol_str<'ast>(data: &'ast (), sym: SymbolId) -> ffi::FfiStr<'ast> {
-    unsafe { as_driver_cx(data) }.symbol_str(sym).into()
+    unsafe { as_driver(data) }.symbol_str(sym).into()
 }
 
 extern "C" fn resolve_method_target<'ast>(data: &'ast (), id: ExprId) -> ItemId {
-    unsafe { as_driver_cx(data) }.resolve_method_target(id)
+    unsafe { as_driver(data) }.resolve_method_target(id)
```
Veetaha, line 198: Technically `unsafe` must be propagated to the surrounding `extern "C"` functions, because they bear an implicit contract that `data` is a wellformed pointer to a `MarkerContextWrapper`. If the functions are not marked as `unsafe` it means you can call the `resolve_method_target`, for example like this from safe Rust and trigger undefined behavior:
```rust
resolve_method_target(&(), some_expr_id)
```
So the unsafe invariant must be propagated to the caller up to that point where safe Rust knows for sure it passes a wellformed pointer to `MarkerContextWrapper` and no other invoking safe Rust code can break that.

marker_adapter/src/context/map.rs
```diff
@@ -0,0 +1,76 @@
...
+    ast::item::{EnumVariant, Field},
+    context::{AstMap, AstMapCallbacks},
+    ffi,
+    lint::Level,
+    prelude::*,
+};
+
+#[repr(C)]
+pub struct AstMapWrapper<'ast> {
+    pub driver: &'ast dyn AstMapDriver<'ast>,
```
Veetaha, line 11: What's the reason to make this pub as well as in `MarkerContextWrapper`?

marker_api/src/context.rs
```diff
@@ -1,21 +1,23 @@
 //! This module is responsible for the [`MarkerContext`] struct and related plumbing.
 //! Items in this module are generally unstable, with the exception of the
 //! exposed interface of [`MarkerContext`].
+//!
+//! Checkout the documentation of `marker_adapter::context` for an explanation
+//! how the backend of these structs are implemented.
```
Veetaha, line 6:
```suggestion
//! Checkout the documentation of `marker_adapter::context` for an explanation
//! of how the backend of these structs is implemented.
```

marker_api/src/context/map.rs
```diff
@@ -0,0 +1,90 @@
...
+        (self.callbacks.item)(self.callbacks.data, id).copy()
+    }
+
+    /// Returns the [`EnumVariant`] belonging to the given [`VariantId`].
+    ///
+    /// This might return [`None`] if the requested [`EnumVariant`] is no longer
+    /// available. This can happen, if the [`EnumVariant`] comes from a different
+    /// crate which is currently not loaded or if the [`VariantId`] has become
+    /// invalid.
+    pub fn variant(&self, id: VariantId) -> Option<&EnumVariant<'ast>> {
```
Veetaha, line 38: All these methods return an `Option`. I suppose the option will be `None` in case of some really hard-to-reproduce bug. I suppose if the `VariantId` was obtained from the current context and wasn't read from some persistent storage or wasn't constructed manually using `unsafe` then this will never return `None`. Maybe it makes sense then to just panic in case `None` is returned complaining that `VariantId` is now dangling and your code contains a bug.

Otherwise what would the use of `marker_api` do when it sees an option here? Is there any possible use case where the user of these method won't unwrap the option? I doubt someone ever needs to check that the option is `None`, and do something else meaningful.

marker_api/src/lint.rs
```diff
@@ -1,23 +1,45 @@
+/// This struct defines a lint.
+///
+/// A struct should never be constructed directly, the [`declare_lint`](crate::declare_lint)
```
Veetaha, line 3:
```suggestion
/// It struct should never be constructed directly, the [`declare_lint`](crate::declare_lint)
```

marker_api/src/lint.rs
```diff
@@ -35,13 +57,11 @@ pub struct Lint {
...
-    // additional fields used  in rustc:
-    // * pub edition_lint_opts: Option<(Edition, Level)>,
-    // * pub future_incompatible: Option<FutureIncompatibleInfo>,
-    // * pub feature_gate: Option<&'static str>,
-    // * pub crate_level_only: bool,
+
+    /// This struct should always be instantiated using the [`declare_lint`](crate::declare_lint)
+    /// macro. This value is simply here, to force any construction to acknowledge the
+    /// instability of manual construction.
+    pub _unstable_i_accept_the_risk_of_instability: u8,
```
Veetaha, line 64: Let's make this zero cost :rocket: 
```suggestion
    pub _unstable_i_accept_the_risk_of_instability: (),
```

marker_rustc_driver/src/context/map.rs
```diff
@@ -0,0 +1,50 @@
...
+use super::RustcContext;
+
+impl<'ast, 'tcx: 'ast> AstMapDriver<'ast> for RustcContext<'ast, 'tcx> {
+    fn item(&'ast self, id: ItemId) -> Option<ItemKind<'ast>> {
+        let rustc_id = self.rustc_converter.to_item_id(id);
+        self.marker_converter.item(rustc_id)
+    }
+
+    fn variant(&'ast self, _id: VariantId) -> Option<&'ast EnumVariant<'ast>> {
+        todo!()
```
Veetaha, line 17: I wonder how the panic across an FFI boundary manifests itself today. Is it going to be debuggable if this todo-ed code is mistakenly invoked during development?

marker_api/src/context.rs
```diff
@@ -83,8 +85,10 @@ where
 /// This context will be passed to each [`LintPass`](`super::LintPass`) call to enable the user
 /// to emit lints and to retrieve nodes by the given ids.
 #[repr(C)]
+#[cfg_attr(feature = "driver-api", derive(typed_builder::TypedBuilder))]
 pub struct MarkerContext<'ast> {
-    driver: &'ast DriverCallbacks<'ast>,
+    ast: AstMap<'ast>,
+    callbacks: MarkerContextCallbacks<'ast>,
```
Veetaha, line 91: So fundamentlly, this changes the call site like this
```diff
- ctx.item(id)
+ ctx.ast().item(id)
```
Not sure how I feel about this. It's just the syntax organizational change. Conceptually, `MarkerContext` stores everything that one needs to query the AST/semantic information, and this separation of `AstMap` adds more boilerplate syntax to access that info.

Although It's hard to suggest better organization here. All I think of is sorting method by AST, semantics, name resolution

marker_adapter/src/context/map.rs
xFrednet: I think it just made it a bit easier than adding a constructor or builder. There is no specific reasoning besides that.

marker_api/src/context/map.rs
xFrednet: I had three reasons:
1. If a user stores the *ID* somewhere and then reloads it or constructs an invalid one
    * I didn't consider that a user would actually not be able to create an ID, without unsafe code :sweat_smile:
2. Rustc compiles each crate individually. During linting it's possible to get the `ItemId` of an item from a dependency (For example by resolving the target of a function call). However, there is no AST representation of the item available, since it's not in the currently compiled crate.
3. If we ever support other drivers, like IDE's where nodes might be deleted again.
    * This one is the least important, as compilers are currently the main target of Marker. And the IDs already have a note in their documentation, that they're not allowed to be stored across `check_*` function boundaries. I considered giving them an `'ast` lifetime, but that sounds like a lot of hassle.

I think 2. is the main reason to keep the result wrapped in an `Option`. We could remove it from the `expr` method, since it's not possible to get an expression id from external expressions.

I'll also add more documentation to `AstMap` to describe when these methods can return `None`.

Rustc uses a general `find() -> Option<Node>` function and then specific `expect_expr()` functions. That could be a model we can easily adapt :thinking: 

Does this reasoning make sense?

marker_rustc_driver/src/context/map.rs
xFrednet: `panic!` calls across FFI boundaries are undefined behavior, in practice they trace to the FFI boundary and then abort with the stack trace until that point. This is something which should be addressed by Marker at some point #10 

Marker has two tests that verify the current behavior, one tracing from the lint crate to Marker and one from the driver. I have to check, why I left this one blank :thinking:

marker_api/src/context.rs
xFrednet: Yeah, AST and semantics are the two things I would probably extract. And, most likely, one that is only for internal callbacks.

One reason for this change is to limit the number of functions on a single struct. Rustc's [TyCtxt](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html#) is a worst-case example of how a context struct can get out of control. It works for rustc and has its reasons for being as it is, but I do find it difficult to navigate occasionally.

The other benefit I see is the option to group documentation. For example, the `AstMap` often returns `Option<Node>`. Here we could just document the reason once and then always just add a reference to the documentation of the main struct instead of having it on every function.

marker_api/src/lint.rs
xFrednet: I believe that `()` inside `#[repr(C)]` structs are undefined behavior. A `&()` reference or pointer, like Marker uses, is fine, but the `()` type isn't AFAIK. Otherwise, this would be the way to go.

marker_adapter/src/context.rs
xFrednet: You might be right on this one. Thinking about it, it would be better to have a dummy struct inside `marker_api`, like `MarkerContextDriverData` and use that for the reference instead. So that would make it `&'ast MarkerContextDriverData` instead of `&'ast ()`. I like this more than the current `&()` one, and it's harder to mess up. In that case, the function wouldn't need to be unsafe either, right :thinking:

marker_api/src/context/map.rs
Veetaha: So, for example, if we lint crate `foo`, then how can we get the `VariantId` from say `std` crate without unsafe read from external storage?

---

Yeah, I agree that for the general use case when users don't try to access the info from the external crate we should make that more convenient to obtain that info (panic by default). I'd make the functions like these targeting for the optimistic case (when users don't access AST from external crates, if that's even possible):
```rust
// Panics (general use case)
fn variant(&self, id: VariantId) -> &EnumVariant<'ast>;

// Returns `None`, but I wonder if this will ever be useful to anyone
fn try_variant(&self, id: VariantId) -> Option<&EnumVariant<'ast>>;
```

xFrednet: Let's say we have the following input:

```
fn main() {
	let x = Some(12)
}
``` 

Then the lint could do this:

```
fn check_stmt(...) {
    if let StmtKind::Let(local) = stmt {
        if let Some(ExprKind::Ctor(ctor)) = local.init() {
            // https://docs.rs/marker_api/latest/marker_api/ast/struct.AstQPath.html
            let path = ctor.path();

            if let AstPathTarget::Variant(var_id) = path.resolve() {
                // Here we have a the valid `VariantId` ID for `Some`.
                //
                // Right now, it would be useless, but at one point, Marker should enable
                // users to request IDs from dependencies and get the semantic variant
                // information, which is actually available
            }
        }
    }
}
```

Having the `try_` prefix might work well :+1:

xFrednet:
Woops, I didn't want to push yet, oh well, doesn't matter. I'll work on the other comments tomorrow! :)

marker_api/src/context/map.rs
xFrednet: I've decided to use the `expect_` prefix instead. IMO it's too easy to get a valid ID for an unavailable AST node, as to make the panicking variant the default.

But there are some nodes, where it's certain, that a given ID also has an AST node. This being bodies, statements, and expressions. For those, I've changed the return type. Beforehand, I was a bit uncertain about this. Statements, for example, can also wrap items, but my current understanding is that this should still be safe.

I'll make the adjustments and then run it on a few test crates, like tokio and bevy.

xFrednet:
Okay, this PR should be ready again :) The CI sadly doesn't run due to the merge conflict, but I triggered it manually, and everything is green :)

marker_api/src/lint.rs
Veetaha: I don't think that `()` is UB in `#[repr(C)]`. If it was so, then `rustc` would definitely not allow using ZST in `#[repr(C)]`, because the paradigm of `rustc` is that everything that may trigger a UB must be explicitly annotated with `unsafe` or otherwise a compile error will be emitted.

According to [rustonomicon](https://doc.rust-lang.org/nomicon/other-reprs.html) ZST are allowed in `#[repr(C)]` and they still ZST (counterintuitive to C's requirement that empty struct must take up 1 byte of memory).

xFrednet: Ah okay, then it was just an empty struct, with only one `_: ()` field in it, that caused a warning. I'll update it :)

marker_api/src/context/map.rs
Veetaha: I'm not sure why `rustc` uses `expect_` prefix, because if you see `expect` you'd likely think there will be an argument for the expect message, but there isn't any.

For example, `Result` uses `unwrap_` prefix for [`unwrap_err`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.unwrap_err), and it seems more logical to me

marker_adapter/src/context/map.rs
Veetaha: Well, if we consider `marker_adapter` to be entirely private to marker, then it's fine, but if there is a chance this may become stable for external users, I'd refrain from exposing a public field

Veetaha:
I think it's fine to do a force-push, if new changes are grouped in an obvious separate commit, so don't bother

marker_adapter/src/context/map.rs
xFrednet: `marker_adapter` is only targeted towards drivers, but if this is a concern, it's easy to just add a constructor. I'll do that in the next commit.

marker_api/src/context.rs
```diff
@@ -308,48 +308,61 @@ impl<'ast> MarkerContext<'ast> {
...
+
+/// This type is used by [`MarkerContextCallbacks`] as the first argument to every
+/// function. For more information, see the documentation of the `data` field
+/// or from `marker_adapter::context`.
+///
+/// This type should never be constructed and is only meant as a pointer
+/// casting target.
+#[repr(C)]
+#[cfg_attr(feature = "driver-api", visibility::make(pub))]
+struct MarkerContextData {
```
Veetaha, line 361: Not sure that having this struct removes the need for callbacks to be unsafe themselves. Because you can still trigger UB from safe Rust, although now only within this crate like this:
```rust
(self.callbacks.resolve_ty_ids)(&MarkerContextData { _data: 0 }, path.into())
```
Anyway, that's not too important to postpone this PR. We can refactor unsafe any time. It still works this way, so that should be fine

marker_api/src/context/map.rs
```diff
@@ -10,9 +10,32 @@ use crate::{
...
+/// Generally speaking, it's advised to use the failable variant of the provided
+/// functions and just bail if the expected node is unavailable. The API also
+/// provides some functions that automatically unwrap the requested nodes. These
+/// functions, starting with `unwrap_`, are great for prototyping but should be
+/// used carefully.
+///
+/// FIXME(xFrednet): While the driver might not have the dependency loaded, it
+/// usually has some semantic information about what types and functions are
+/// available. The API should provide some way to request this semantic information
+/// based on the ID. (See: rust-marker/marker#266)
```
Veetaha, line 32: Maybe we should move this FIXME comment outside of the docs and use `//` for it?

Or maybe remove the `FIXME` part and make it a natural part of the doc comment like this.
```suggestion
/// # Future plans
/// While the driver might not have the dependency loaded, it usually has some semantic information 
/// about what types and functions are available. The API should provide some way to request this 
/// semantic information based on the ID. (See: https://github.com/rust-marker/marker/issues/266).
```
Otherwise `FIXME` sound like a comment that unintentionally percolated to the docs

marker_rustc_driver/src/conversion/marker/item.rs
```diff
@@ -115,6 +115,13 @@ impl<'ast, 'tcx> MarkerConverterInner<'ast, 'tcx> {
                             variant.disr_expr.map(|anon| self.to_const_expr(anon)),
                         )
                     }));
+                    // A block to limit the lifetime of the mut borrow
+                    {
+                        let mut cache = self.variants.borrow_mut();
+                        for var in variants {
+                            cache.insert(var.id(), var);
+                        }
```
Veetaha, line 123: If you use `extend` instead of a for loop and the iterator returns the expected number of elements in the `size_hint()` method, then the map can pre-allocate enough space beforehand, meaning that `extend` is always faster than a for loop with `insert`, because it has more context, and won't do repeated allocations and even additional checks for growing the backing storage.
```suggestion
                        self.variants.borrow_mut().extend(variants.into_iter().map(|var| (var.id(), var)));
```

Same allies to all other for/insert loops added in the diff

Veetaha (approved):

marker_api/src/context.rs
xFrednet: It doesn't remove the potential for UB, but makes it way less likely, since the type is private to the module. So, this is a slight improvement. We can add unsafe later :)

marker_rustc_driver/src/conversion/marker/item.rs
xFrednet: Yeah, you're absolutely right. There should be a Clippy lint for this ^^

xFrednet:
I've squashed the previous commits and the two comments. (The functions are still not unsafe, that can be done at a later point :))

xFrednet:
The semver check fails are all expected:
1. for the new field in `Lint`
2. moved functions from `MarkerContext` -> `AstMap`.

Veetaha (approved):

Issue #120, opened 2023-02-27
Test for API breaking changes in CI


xFrednet:
Currently, the CI only checks that the `marker_api` crate can be compiled on stable. It's up to the reviewer and programmer to ensure that the API doesn't break. This is fine right now, as the API is under active development and still changes frequently. However, the goal of Marker is to create a stable API.

The rust ecosystem has tools to prevent accidental public API changes, like [cargo-semver-checks](https://crates.io/crates/cargo-semver-checks). Once v0.1.0 or a higher version has been released, I think it would be good to utilize such a tool.

This issue is blocked for now until v0.1.0 or a later point, when we've defined what stability means for marker and ii API breakage will be allowed under specific circumstances.

xFrednet:
I've done a bit of testing on this, but nothing extensive. One thing I noticed, is that by default, it compares the API to the version from crates.io. This means that once the API is broken on master, every PR would fail this check. Instead, the semver check should check if the changes of the PR add new breaking changes, in comparison to master.

During initial development, `v0.*` API breakage is allowed. During this time, I want to have this check, to ensure that we don't break the API unintentionally and that these changes are clearly tracked in our changelog.

PR #262, opened 2023-09-28
CI: Add semantic versioning test for controlled breakage

xFrednet:
This PR adds [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) to marker's CI, as a sanity check. This CI job shouldn't block a PR from being merged, but it's a good indicator, that a change should be listed in Marker's changelog.

Ideally, I would have liked a job status, between success and failed. Maybe a nice orange icon, saying that the job failed, but that that's acceptable? I know GitLab had something like this in the past, but GitHub doesn't seem to support it. It's either: success, failure, or cancelled. Oh well, an indicator is an indicator.

Here is an example, of a failed semver check, due to a removed public function: https://github.com/xFrednet/marker/actions/runs/6340462879/job/17221897343?pr=1

---

I should really stop procrastinating on my uni assignments... Oh well, at least this way I have the feeling I'm doing something useful.

Anyways:

Closes: rust-marker/marker#120

And anyone reading this, have a beautiful day :D

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,12 +140,7 @@ jobs:
   # This check is allowed to fail, while Marker is still unstable. It's mostly
   # a sanity check, to make sure we don't have any unintentional API breakage :)
   semver-checks:
-    env:
-      PR_NUM: ${{ github.event.pull_request.number }}
     runs-on: ubuntu-latest
-    # This CI check isn't required to pass just yet.
-    # Marker isn't stable so semver breaking changes are allowed.
-    continue-on-error: true
     if: ${{ github.event.pull_request }}
     steps:
       - uses: actions/checkout@v4
@@ -165,6 +160,6 @@ jobs:
           --baseline-root './upstream'
           --default-features
           --release-type minor
-      - run: gh pr edit $PR_NUM --add-label "I-api-break"
-        if: ${{ github.repository == 'rust-marker/marker' && failure() }}
-      - run: echo "$PR_NUM"
+        # This CI check isn't required to pass just yet.
+        # Marker isn't stable so semver breaking changes are allowed.
+        continue-on-error: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,23 +140,31 @@ jobs:
   # This check is allowed to fail, while Marker is still unstable. It's mostly
   # a sanity check, to make sure we don't have any unintentional API breakage :)
   semver-checks:
+    env:
+      PR_NUM: ${{ github.event.pull_request.number }}
     runs-on: ubuntu-latest
+    # This CI check isn't required to pass just yet.
+    # Marker isn't stable so semver breaking changes are allowed.
+    continue-on-error: true
     if: ${{ github.event.pull_request }}
     steps:
       - uses: actions/checkout@v4
         with:
           ref: ${{ github.base_ref }}
-          path: './base'
+          path: './upstream'
       - uses: actions/checkout@v4
         with:
-          path: './source'
-      - run: source/scripts/download/cargo-semver-checks.sh
+          path: './downstream'
+      - run: downstream/scripts/download/cargo-semver-checks.sh
       - run: >-
           cargo semver-checks
-          --manifest-path './source/Cargo.toml'
+          --manifest-path './downstream/Cargo.toml'
           --package marker_api
           --package marker_utils
           --package marker_uitest
-          --baseline-root './base'
+          --baseline-root './upstream'
           --default-features
           --release-type minor
+      - run: gh pr edit $PR_NUM --add-label "I-api-break"
+        if: ${{ github.repository == 'rust-marker/marker' && failure() }}
+      - run: echo "$PR_NUM"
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -135,3 +136,27 @@ jobs:
...
+  semver-checks:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.base_ref }}
+          path: './base'
+      - uses: actions/checkout@v4
+        with:
```
xFrednet, line 151:
```
with:
    ref: ${{ github.head_ref }}
```
Sadly doesn't work. I'm assuming that just specifying the ref, is not enough, and I would also need to change the remote to the fork, where the branch is from? Anyways, just leaving that part out, does the trick :)

.github/workflows/ci.yml
```diff
@@ -135,3 +136,27 @@ jobs:
 
       # Check that the release automation works as expected
       - run: scripts/release/test.sh
+
+  # This check is allowed to fail, while Marker is still unstable. It's mostly
+  # a sanity check, to make sure we don't have any unintentional API breakage :)
+  semver-checks:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request }}
```
Veetaha, line 144: Let's add [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error), so that if this job fails the workflow will still be displayed as green in the Actions tab of the repo.
```suggestion
    runs-on: ubuntu-latest
    # This CI check isn't required to pass just yet.
    # Marker isn't stable so semver breaking changes are allowed.
    continue-on-error: true
    if: ${{ github.event.pull_request }}
```

.github/workflows/ci.yml
```diff
@@ -135,3 +136,27 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.base_ref }}
+          path: './base'
```
Veetaha, line 149: Maybe `upstream/downstream` could be better names rather than `base/source`?

Veetaha (approved):
Nice quality of life improvement :+1:

.github/workflows/ci.yml
xFrednet: This will require us to always look at the log, to see if the semver-check actually fails. :thinking:

For a second, I contemplated if it would be cool to automatically add/remove the `I-api-break` label. But then I considered the implications of giving PR edit permissions to a normal user editable CI job, and decided against it...

xFrednet: I think having the check green by default is better than having it red :+1: 

Do you have another idea, how we could maybe get notified about the API breakage, without checking the job every time?

I considered checking for the `I-api-break` label and error if that one isn't present. However, the CI would not rerun if the label is added afterwards, resulting in a red CI again :/

The last test commit, contains some API breakage, just for a reference how it would look like.

Veetaha: I wouldn't put the `continue-on-error` on the `cargo semver-checks` step itself.
This way the check itself will always be green, which I suppose isn't what we want. It should be read if a major breaking change was detected.

The `continue-on-error` at the job level makes it so that the workflow itself is displayed as green in the github action runs like here: https://github.com/rust-marker/marker/actions
image

But inside the PR view the job should be displayed in red.

---

Regarding the notification. Yes! Auto-labeling the PR without granting the downstream fork broad write permissions is doable. Checkout out this [blog post](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

I think we can use the [`workflow_run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) approach where the untrusted regular CI on `pull_request` event runs the semver checks, and in case the check detects a breaking change the CI job should mark itself as failed. Then, a followup privileged CI is triggered on `workflow_run` event. It will run in the context of the upstream repo (and not PR merge commit), so it will not observe any PR code modifications, and thus is safe. In our case it doesn't even need to check out the PR code or even do any code checkout whatsoever.

The unprivilleged `semver-checks` job needs to upload an artifact with the results of the checks, and the privileged workflow downloads the artifact ([docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow)) and make take advantage from repo write permissions.

In our case the artifact should store the PR number and the boolean result of the check so that the workflow could determine if it needs to add a PR label (if one isn't already there), or remove it (if there is one).

xFrednet: Having automated GitHub annotations would be the dream! I've subscribed to the state of the issue now, but it's likely that that solution will take its time.

I've now moved the [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error), what I dislike a bit, is that it marks the PR as "failed". The CI job itself was less my concern.

I think the split CI approach, with one privileged CI for applying the label, would work well for the meantime. I'll try to find some time to test this.

Thank you for the suggestion!

Veetaha:
I think we may just merge this as is, and the auto-tagging workflow separately. It's better to have something than nothing, right?

xFrednet:
You're right, I have to learn to get a better balance of what to merge or what to perfect first :)

xFrednet:
I accidentally committed the breaking change as well :facepalm:

PR #263, opened 2023-09-28
uitest: Update `ui_test` from `v0.11.7` to `v0.21.2` and other deps

xFrednet:
Not much to say. I just checked crates.io and updated every dependency I could. Then I ran `cargo update` and everything was done.

Updating `ui_test` was a bin painful due to the big version difference, but it seems to be working now :D

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,7 @@ jobs:
           --workspace
           --exclude marker_rustc_driver
           --exclude marker_lints
+          --exclude marker_uitest
 
   # Check for unused dependencies that uses simple regex search,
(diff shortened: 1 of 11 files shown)


xFrednet:
*ci / rust-unused-dependencies* Just caught its first unused dependency :tada:

xFrednet:
Clap or at least a dependency of clap has increased the minimum rust version to Rust 1.70. I think at one point we should have a discussion how we handle the minimal rust version of Marker and its dependencies. This is just frustrating rn

xFrednet:
Damn, dealing with dependencies can take some time. I thought this would just take 10 minutes ^^ Anyways it seems to be working now.

I think I can just merge this PR directly, since it's only a dependency update :)

PR #194, opened 2023-07-23
API: Remove `CallableData` from `FnItem` and use fields instead

xFrednet:
cc: rust-marker/marker#181

PR #265, opened 2023-09-30
API: Removed the `CallableData` trait

xFrednet:
The `CallableData` trait, was designed with the idea to have a general trait for function like things, including `FnItem`, `ClosureExpr`, `SenFnPrtTy`. However, after working with Marker for a bit, I learned that these AST nodes are just too different. At least, too different, to have the `CallableData` trait as the main API.

I'm also not sure what the use case would be, where a syntactic type should be handled the same way as a `FnItem`.

This PR simply removes the trait and moves the remaining functions to the `SenFnPrtTy` type. https://github.com/rust-marker/marker/pull/194 already removed the trait from `FnItem` and `ClosureExpr`.

This PR also deletes the `SynClosureTy` since it's not actually a syntactic type that can be written ^^


marker_api/src/ast/ty/fn_ty.rs
```diff
@@ -30,26 +28,6 @@ impl<'ast> SemFnTy<'ast> {
     }
 }
 
-/// The syntactic representation of a
-/// [closure type](https://doc.rust-lang.org/reference/types/closure.html).
```
Veetaha, line 34: That's weird how there could be a syntactic representation of a closure type because there is no syntax to denote a type of the closure except for `impl Fn*`. However,  `impl Fn*` should not be coupled with closures, that's a broader concept.

So I wonder what this struct represented then

Veetaha (approved):

marker_api/src/ast/ty/fn_ty.rs
xFrednet: I don't actually know anymore why I added this struct. My guess is that this is a left over from an old architecture. There was a time, where I planned to merge both syntactic and semantic types into a single representation. My idea was that this would make it easier, to work with types. Then you could compare syntactic and semantic types etc.

However, while working on semantic types, I quickly noticed that these are way more different than I thought. It also became clear that this combination will most likely be painful for everyone involved, with very few benefits.

During the move to split semantic and syntactic types, I just added the `Syn` prefix to most types. The struct might have been intended to be a semantic type, but got caught in the crossfire.

PR #269, opened 2023-10-03
uitest: Don't bless by default

xFrednet:
The `ui_test` update from rust-marker/marker#263 changed the process, to bless by default. This was unexpected for me, and can lead to users thinking that a test passed, even if it should have failed.

See: https://github.com/rust-marker/marker/pull/268#discussion_r1343790060 as an example

### DIFF
--- a/marker_lints/tests/uitest.rs
+++ b/marker_lints/tests/uitest.rs
@@ -6,19 +6,8 @@ fn main() -> color_eyre::Result<()> {
 
     config.dependencies_crate_manifest_path = Some("./Cargo.toml".into());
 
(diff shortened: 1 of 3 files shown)


Veetaha (approved):

xFrednet:
Okay, I'm done for today. I just can't figure out how to fix the path replacement for windows :sob:

Veetaha:
I don't know why it works like that on windows, but you can unblock this with these two filters only. Tested it on my windows laptop
```rust
    config.filter(r"\\/", r"/");
    config.filter(r"\\\\", r"/");
```

xFrednet:
A savior as always, thank you soo much!!! :heart: This has been so frustrating!

PR #271, opened 2023-10-04
Rustc: Librarify `marker_rustc_driver`

### LINKED ISSUE #257: `marker_rustc_driver` as a library?
How would you feel about exposing `marker_rustc_driver` internals as a library?

Context: I have been experimenting with idea of running Marker lints under a Dylint driver, and I have a [PoC](https://github.com/smoelius/dylint/tree/marker/examples/testing/marker). The PoC is essentially a single [configurable](https://github.com/trailofbits/dylint#configurable-libraries) Dylint library. The configuration (i.e., `dylint.toml` file) tells the Dylint library where to find the Marker libraries. For example, in my tests, the file looks as follows:

```toml
[[marker.lint_crates]]
name = "marker_lints"
path = "/tmp/.../target/marker/lints/libmarker_lints.so"
```
So, essentially, the PoC is a shared library that loads other shared libraries. For anyone interested in trying it out, the following should work:
```
git clone https://github.com/smoelius/dylint --branch marker
cd dylint/examples/testing/marker
cargo test
```
This will run a slightly modified version of the `diag_msg_uppercase_start` tests using the setup described above.

---

To make the above work, I had to copy a [bunch of `marker_rustc_driver`'s code](https://github.com/smoelius/dylint/blob/9877452fdbfff5de1a659e11d28ffa3613de528b/examples/testing/marker/src/lib.rs#L23-L29). I would like to include this PoC as a test in the Dylint repository, but I would prefer to not copy that code (hence, this question).

Note that GitHub access to the library would suffice. So if you do not want to publish `marker_rustc_driver` internals as a library on crates.io (which would be perfectly reasonable), there may be a way to expose the internals as a library through conditional compilation.

smoelius:
Fixes #257

How do you feel about this?

I would summarize the changes as follows:
- marker_rustc_driver/src/bin/marker_rustc_driver.rs is a new file that contains a slightly modified version of the old `main` function.
- `try_main` was updated to accept an `impl Iterator<Item = String>`. `main` passes `std::env::args()` as this argument.[^1^]
- The following were made `pub`:
  - `try_main`
  - `lint_pass`
  - `MainError`
  - `process_crate`

---

[^1^]: This change was not strictly necessary. But since `try_main` is now a public function, the idea of it reading from `std::env::args()` directly seemed weird to me.

### DIFF
--- a/marker_rustc_driver/src/bin/marker_rustc_driver.rs
+++ b/marker_rustc_driver/src/bin/marker_rustc_driver.rs
@@ -1,29 +1,12 @@
 #![doc = include_str!("../../README.md")]
 #![feature(rustc_private)]
 #![feature(lint_reasons)]
-#![feature(non_exhaustive_omitted_patterns_lint)]
 #![warn(clippy::pedantic)]
-#![warn(non_exhaustive_omitted_patterns)]
 #![allow(clippy::missing_panics_doc)]
-#![allow(clippy::module_name_repetitions)]
-#![allow(clippy::needless_lifetimes, reason = "lifetimes will be required to fix ICEs")]
-#![allow(clippy::needless_collect, reason = "false positives for `alloc_slice`")]
-#![allow(clippy::too_many_lines, reason = "long functions are unavoidable for matches")]
 
-extern crate rustc_ast;
-extern crate rustc_data_structures;
 extern crate rustc_driver;
-extern crate rustc_errors;
-extern crate rustc_hash;
-extern crate rustc_hir;
-extern crate rustc_hir_analysis;
-extern crate rustc_interface;
-extern crate rustc_lint;
-extern crate rustc_lint_defs;
-extern crate rustc_middle;
 extern crate rustc_session;
 extern crate rustc_span;
-extern crate rustc_target;
 
 use std::env;
 
--- a/marker_rustc_driver/src/bin/marker_rustc_driver.rs
+++ b/marker_rustc_driver/src/bin/marker_rustc_driver.rs
@@ -1,0 +1,68 @@
+#![doc = include_str!("../../README.md")]
+#![feature(rustc_private)]
+#![feature(lint_reasons)]
+#![feature(non_exhaustive_omitted_patterns_lint)]
+#![warn(clippy::pedantic)]
+#![warn(non_exhaustive_omitted_patterns)]
+#![allow(clippy::missing_panics_doc)]
+#![allow(clippy::module_name_repetitions)]
+#![allow(clippy::needless_lifetimes, reason = "lifetimes will be required to fix ICEs")]
+#![allow(clippy::needless_collect, reason = "false positives for `alloc_slice`")]
+#![allow(clippy::too_many_lines, reason = "long functions are unavoidable for matches")]
+
+extern crate rustc_ast;
+extern crate rustc_data_structures;
+extern crate rustc_driver;
+extern crate rustc_errors;
+extern crate rustc_hash;
+extern crate rustc_hir;
+extern crate rustc_hir_analysis;
+extern crate rustc_interface;
+extern crate rustc_lint;
+extern crate rustc_lint_defs;
+extern crate rustc_middle;
+extern crate rustc_session;
+extern crate rustc_span;
+extern crate rustc_target;
+
+use std::env;
+
+use rustc_session::config::ErrorOutputType;
+use rustc_session::EarlyErrorHandler;
+
+use marker_rustc_driver::{try_main, MainError};
+
+const BUG_REPORT_URL: &str = "https://github.com/rust-marker/marker/issues/new?template=panic.yml";
+
+fn main() {
+    let handler = EarlyErrorHandler::new(ErrorOutputType::default());
+    rustc_driver::init_rustc_env_logger(&handler);
+
+    // FIXME(xFrednet): The ICE hook would ideally distinguish where the error
+    // happens. Panics from lint crates should probably not terminate Marker
+    // completely, but instead warn the user and continue linting with the other
+    // lint crate. It would also be cool if the ICE hook printed the node that
+    // caused the panic in the lint crate. rust-marker/marker#10
+
+    rustc_driver::install_ice_hook(BUG_REPORT_URL, |handler| {
+        handler.note_without_error(format!("{}", rustc_tools_util::get_version_info!()));
+        handler.note_without_error("Achievement Unlocked: [Free Ice Cream]");
+    });
+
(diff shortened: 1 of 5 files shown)


smoelius:
I think this may be the happiest I can get CI?

xFrednet:
The CI has some tests, that will fail until v0.3.0 has been released. I'll review this during the weekend :)

marker_rustc_driver/src/bin/marker_rustc_driver.rs
```diff
@@ -0,0 +1,68 @@
...
+extern crate rustc_hash;
+extern crate rustc_hir;
+extern crate rustc_hir_analysis;
+extern crate rustc_interface;
+extern crate rustc_lint;
+extern crate rustc_lint_defs;
+extern crate rustc_middle;
+extern crate rustc_session;
+extern crate rustc_span;
+extern crate rustc_target;
```
xFrednet, line 26:
```suggestion
#![doc = include_str!("../../README.md")]
#![feature(rustc_private)]
#![feature(lint_reasons)]
#![warn(clippy::pedantic)]
#![allow(clippy::missing_panics_doc)]

extern crate rustc_driver;
extern crate rustc_session;
extern crate rustc_span;
```

This section only needs to contain the features and `extern crates`, needed in this file. I think it's better to reduce the number significantly.

I would also like to have a doc comment here explaining, that this is just the binary, that uses the `lib.rs` file. Just to make it easier for newcomers to navigate.

marker_rustc_driver/src/bin/marker_rustc_driver.rs
```diff
@@ -0,0 +1,68 @@
+#![doc = include_str!("../../README.md")]
```
xFrednet, line 1: Today I learned that you can have multiple binaries simply by adding a `bin` folder under source o.O

Cargo's documentation states:

> * The default executable file is `src/main.rs`.
>   * Other executables can be placed in `src/bin/`.

Could you move this file to `src/main.rs`? For me, that's the default location where I would look first, and it's also the default executable, which should be located in `src/main.rs` according to the docs.

I assume, this was just a stylistic choice, as moving it to `src/main.rs` seems to work the same way. If there is another reason, please share it :)

xFrednet (commented):
I think these changes are very reasonable. And I'm okay with merging it, once my comments have been addressed :)

Note, that I've just updated the used nightly version and the CI should be fully functional now. It would be good if you could rebase. And thank you for waiting on my review :D

marker_rustc_driver/src/bin/marker_rustc_driver.rs
smoelius: > I assume, this was just a stylistic choice

To be honest, I didn't know a package could have both a `src/lib.rs` and a `src/main.rs`. So, TIL too. :)

smoelius:
@xFrednet I think I've addressed all of your comments. Please let me know if you see anything else.

smoelius:
Sorry, I noticed a typo. I think the PR should be good now.

xFrednet:
This version looks good to me. I have one question, and then we should be able to merge this. The changes make `try_main` public. Will dylint call this function or is this just for Markers binary?

smoelius:
> The changes make `try_main` public. Will dylint call this function or is this just for Markers binary?

`try_main` is just for Marker's binary.

Right now, the PoC uses [just these exports](https://github.com/trailofbits/dylint/blob/81abbe57948b24d78e0d5012c234eeca970913bb/examples/testing/marker/src/lib.rs#L10-L11):
```rust
use marker_adapter::{Adapter, LintCrateInfo};
use marker_rustc_driver::lint_pass;
```
I split-out the library at `try_main` because (1) it was easy to do, and (2) it was easy to see what changed.

But from my perspective, the split does not have to occur at `try_main`.

(No worries if you would like additional changes.)

xFrednet:
My question about `try_main` was just to ensure you're not running the compiler twice :)

marker_rustc_driver/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+fn register_tracked_env(sess: &mut rustc_session::parse::ParseSess, vars: &[(&'static str, String)]) {
+    use rustc_span::Symbol;
+    let env = sess.env_depinfo.get_mut();
+
+    for (key, value) in vars {
+        env.insert((Symbol::intern(key), Some(Symbol::intern(value))));
+    }
+}
+
+fn register_tracked_files(sess: &mut rustc_session::parse::ParseSess, lint_crates: &[LintCrateInfo]) {
```
xFrednet, line 117: Rustc uses caching where possible. This function, and the function above, register relevant files and environment values which should be checked, when rustc decides if a cache can be reused or not.

This one mainly registers the `Cargo.toml` file, as that's where Marker's lints are configured and the lint crate files. It basically just ensures that we don't use caches if any lint crate was updated. Do you maybe want to make this one public as well? I'm not sure how dylint handles this kind of thing.

This is just a suggestion, if you handle this differently or don't need it, I'm happy to merge this as is :)

smoelius: > Do you maybe want to make this one public as well?

It's kind of you to ask, but I don't think it's necessary.

If "`marker_rustc_driver` as a library" feels "incomplete" to you without this function, I can sympathize with that feeling. But I would recommend to keep the interface as small as possible until me or someone else complains. :)

xFrednet: I just wanted to mention it, since i[#294]t was a thing we ran into after some testing :)

xFrednet (approved):

marker_rustc_driver/src/lib.rs
smoelius: I'm curious now (because caching has been the source of several bugs in Dylint). Is there a direct connection between caching and #294?

xFrednet: The reference to #294 seems to be an accidental copy-pasta :spaghetti:. Sorry, about that.

There is a direct connection between the function I marked here and caching. This is the PR that prevented caching for Marker: https://github.com/rust-marker/marker/pull/211

Do you have some examples for caching related bugs in Dylint and how you solved them?

smoelius: > Do you have some examples for caching related bugs in Dylint and how you solved them?

Two that come to mind are:
- https://github.com/trailofbits/dylint/issues/650
  - fixed by: https://github.com/trailofbits/dylint/pull/813
- https://github.com/trailofbits/dylint/issues/856
  - fixed by: https://github.com/trailofbits/dylint/pull/866

Issue #275, opened 2023-10-07
[Feat]: Make it possible to use `AstPathSegment` and `Ident` as an emission node


Veetaha:
### Summary

I tried to update my `ToStringOnCowStr` lint to 0.3.0, but I couldn't make it work seamlessly as I wanted.

Here is the code that I have and I want to focus on:

```rust
ctx.emit_lint(
    TO_STRING_ON_COW_STR,
    method,
    "replace this with .into_owned()",
)
.span(method.method().ident().span());
```

And here is the full code for more context:

<details>
<summary>Details</summary>

```rust
use marker_api::prelude::*;
use marker_api::{LintPass, LintPassInfo, LintPassInfoBuilder};

#[derive(Default)]
struct ToStringOnCowStrPass {}
marker_api::export_lint_pass!(ToStringOnCowStrPass);

marker_api::declare_lint! {
    /// # What it does
    /// This lints detects the usage of `.to_string()` on a `Cow<str>`.
    ///
    /// # Example
    /// ```rs
    /// PathBuf::from("foo").to_string_lossy().to_string()
    /// ```
    ///
    /// Use instead:
    /// ```rs
    /// PathBuf::from("foo").to_string_lossy().into_owned()
    /// ```
    ///
    /// If you intended to clone the `Cow`, then use `.clone()` or `.to_owned()`.
    TO_STRING_ON_COW_STR,
    Deny,
}

impl LintPass for ToStringOnCowStrPass {
    fn info(&self) -> LintPassInfo {
        LintPassInfoBuilder::new(Box::new([TO_STRING_ON_COW_STR])).build()
    }

    // Comment here
    fn check_expr<'ast>(
        // comment there
        &mut self,
        ctx: &'ast MarkerContext<'ast>,
        expr: ExprKind<'ast>,
    ) {
        let ExprKind::Method(method) = expr else {
            return;
        };

        if method.method().ident().name() != "to_string" {
            return;
        }

        let sem::TyKind::Adt(reciever) = method.receiver().ty() else {
            return;
        };

        if !ctx
            .resolve_ty_ids("std::borrow::Cow")
            .contains(&reciever.def_id())
        {
            return;
        }

        let [sem::GenericArgKind::Ty(str)] = reciever.generics().args() else {
            return;
        };

        if !matches!(str, sem::TyKind::Text(str) if str.is_str()) {
            return;
        }

        ctx.emit_lint(
            TO_STRING_ON_COW_STR,
            method,
[...]

PR #278, opened 2023-10-07
Use `HasSpan` trait more extensively for more ergonomic API

Veetaha:
I added `HasSpan` trait to all types that had a custom `span()` method not included in that trait. I also replace `&Span` with `impl HasSpan` in all public methods in `marker_api` to allow passing nodes to them directly.

I even added a new lint to `marker_lints` that should catch the cases when `impl HasSpan` isn't used in a public method, but only after I implemented it I noticed that item visibility information isn't availble in `marker_api` yet. That's a bummer, because this custom lint doesn't make sense for non-pub methods and emits a lot of false positivies due to this fact. Anyway, since #26 has to be eventually implemented I left the new custom lint in code, and added ignores at crate-level for it.


marker_lints/src/not_using_has_span_trait.rs
```diff
@@ -0,0 +1,44 @@
...
+
+marker_api::declare_lint! {
+    /// ### What it does
+    /// Suggests using `impl HasSpan` for functions that need to take `Span` as
+    /// a parameter to make them more ergonomic.
+    NOT_USING_HAS_SPAN_TRAIT,
+    Warn,
+}
+
+pub(crate) fn check_item<'ast>(cx: &'ast MarkerContext<'ast>, item: ItemKind<'ast>) {
```
Veetaha, line 11: I'm struggling with code organization here. Since there is a single lint pass for the entire crate, then if I want to define unrelated lints in separate modules I can't write their code inside of a trait impl. Writing code inside of a trait impl is cool because I can leverage rust-analyzer to paste the stub trait impl and let me fill it in. With the free function I had to copy the signature from the trait's declaration manually. I wish we could improve this situation.

marker_lints/src/not_using_has_span_trait.rs
```diff
@@ -0,0 +1,44 @@
...
+        )
+        .span(param.ty().span());
+    }
+}
+
+fn strip_references(ty: ast::TyKind<'_>) -> ast::TyKind<'_> {
+    match ty {
+        ast::TyKind::Ref(ref_ty) => strip_references(ref_ty.inner_ty()),
+        _ => ty,
+    }
```
Veetaha, line 43: I wonder if this function could be moved to the public API of `TyKind`.

marker_lints/src/not_using_has_span_trait.rs
```diff
@@ -0,0 +1,44 @@
...
+        let ident = path
+            .path()
+            .segments()
+            .last()
+            .unwrap_or_else(|| panic!("BUG: path must have at least one segment, but got: {path:#?}"))
+            .ident();
+
+        if ident != "Span" {
+            continue;
+        }
```
Veetaha, line 28: I didn't find a way to resolve the semantic information of an `ast::TyKind`. That's why I had to use a heuristic that compares the last segment of the path with `Span` which is good enough, but definitely not ideal in cases when there are multiple crates with multiple types named `Span`.

marker_lints/src/not_using_has_span_trait.rs
```diff
@@ -0,0 +1,44 @@
...
+
+        cx.emit_lint(
+            NOT_USING_HAS_SPAN_TRAIT,
+            func,
+            "use impl HasSpan instead of Span for more flexibility",
+        )
+        .span(param.ty().span());
```
Veetaha, line 35: Same problem as in https://github.com/rust-marker/marker/issues/275. I couldn't use the `param.ty()` as the emission node directly, unfortunately.

marker_rustc_driver/src/main.rs
```diff
@@ -15,6 +15,14 @@
 #![allow(clippy::needless_lifetimes, reason = "lifetimes will be required to fix ICEs")]
 #![allow(clippy::needless_collect, reason = "false positives for `alloc_slice`")]
 #![allow(clippy::too_many_lines, reason = "long functions are unavoidable for matches")]
+// Custom Marker lints
+#![cfg_attr(
+    marker,
+    feature(register_tool),
+    register_tool(marker),
+    // This lint applies only to public `marker_api`
+    allow(marker::not_using_has_span_trait)
```
xFrednet, line 24: It might be better, to have the lint allow-by-default and then add the `warn(marker::not_using_has_span_trait)` to `marker_api` instead. I also thought about adding a lint crate for internal marker lints. :thinking:

marker_lints/src/not_using_has_span_trait.rs
xFrednet: Yes, I plan to make the ast types emission nodes as well :)

marker_lints/src/not_using_has_span_trait.rs
xFrednet: it would definitely be a good function to have. [peel_refs](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.Ty.html#method.peel_refs) from rustc might be a good example as well

marker_lints/src/not_using_has_span_trait.rs
xFrednet: Jep, you're right. There should be an option to check this using [`resolve_ty_ids`](https://docs.rs/marker_api/latest/marker_api/context/struct.MarkerContext.html#method.resolve_ty_ids), but there seems to be no way of getting the `TyDefId`, woops :sweat_smile:

marker_lints/src/not_using_has_span_trait.rs
xFrednet: Yeah, I'm currently experimenting a bit in https://github.com/rust-marker/marker-example-lints. Maybe having something like the inventory crate, that you talked about previously could be a good idea :thinking:

marker_api/src/lib.rs
```diff
@@ -7,10 +7,21 @@
 #![allow(clippy::missing_panics_doc)] // Temporary allow for `todo!`s
 #![allow(clippy::new_without_default)] // Not very helpful as `new` is almost always cfged
 #![cfg_attr(not(feature = "driver-api"), allow(dead_code))]
+#![cfg_attr(
+    marker,
+    feature(register_tool),
+    register_tool(marker),
+    // Allowed because it is triggered on a bunch of private/pub(crate) methods.
+    // There isn't a way to inspect the item visibility in this lint's impl yet.
+    // Once rust-marker/marker#26 is done and lint impl ignores private functions
```
xFrednet, line 16: Yeah, that issue is becoming a real blocker. I've added it to the v0.4.0 milestone and plan to work on it soon :)

marker_api/src/ast/ty.rs
```diff
@@ -182,7 +182,9 @@ macro_rules! impl_ty_data {
             fn as_kind(&'ast self) -> $crate::ast::ty::TyKind<'ast> {
                 self.into()
             }
+        }
 
+        impl<'ast> $crate::span::HasSpan<'ast> for $self_ty {
             fn span(&self) -> &$crate::span::Span<'ast> {
                 $crate::context::with_cx(self, |cx| cx.span(self.data.span))
             }
```
xFrednet, line 190:
```suggestion
        $crate::span::impl_has_span_via_field!($self_ty, data.span);
```

xFrednet (approved):
Looks good to me, a few nits and comments, but that's it. It's nice to see another lint implementation. It's a good way to find issues and to prioritize them :+1: 

> When adding a new custom lint to marker_lints I also moved it under a module not_using_has_span_trait and named the lint variable just LINT. I did the same with the existing lint for uppercase letter in diagnostic message, and found out that it generated a name conflict. The error I saw due to this came from the internal panic in rustc_lint or some other internal rusrc crate. It was so verbose, it included a backtrace, and looked so scary that I improved the error handling due the lint crates initialization.

This is something which I've been thinking about. Ideally, I would like to have the lint crate in front, instead of `marker`. `allow(marker_lints::XYZ)` makes more sense, and prevents collisions by default. But the chances of this happening, with the previous state of the discussion about tool name spaces is basically zero. We could add something after the `marker` prefix, `marker::<lint_crate>::<lint_name>` or so. Here, I fear that this is too verbose. :thinking:  I'm also not sure if rustc would accept this correctly, but if this is something we want, we could probably convince the compiler team, to accept these changes :) 

Edit: It seems like rustc accepts lint path with multiple segments, so that could be an option ([Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=64de8d22a8420c717e664d1f7be39325)) :+1: This doesn't solve the issue, if multiple lints have the same name in one lint crate, but at least prevents collisions between different ones

---

For now, I like your solution a lot :D

marker_rustc_driver/src/main.rs
Veetaha: Good idea 💡, I forgot about that

marker_lints/src/not_using_has_span_trait.rs
Veetaha: Cool 🤙, I added the same `peel_refs` method to `ast::TyKind` and `sem::TyKind`.

marker_lints/src/not_using_has_span_trait.rs
Veetaha: The problem with `resolve_ty_ids` is that it takes a stringified path as a parameter. It also returns a list of matching types.
In this case we should have a method that takes an `ast::TyKind` as a parameter and returns just a single `sem::TyKind`. I suppose it shouldn't be ambiguous, and only a single type can be resolved given the context of the crate and module where the `ast::TyKind` resides it should always resolve to a single type, right?

marker_lints/src/not_using_has_span_trait.rs
Veetaha: Yeah, I already forgot about `inventory`. I'm thinking what the right API should be though, it's hard to to come up with something both succinct but flexible.

Veetaha: I see that `clippy` uses a single lint pass per lint.
For example `MultiAssignments` is defined in a separate file with its own lint pass ([code](https://github.com/rust-lang/rust-clippy/blob/33f49f33d6f8cec9a5772bc7f3a4e5f1de207b30/clippy_lints/src/multi_assignments.rs#L32)).

And then that lint pass is registered manually ([code](https://github.com/rust-lang/rust-clippy/blob/33f49f33d6f8cec9a5772bc7f3a4e5f1de207b30/clippy_lints/src/lib.rs#L1010)).

That second part of manual registration can be done by `inventory` though.

But let's not forget that lints should be configurable, like [this one](https://github.com/rust-lang/rust-clippy/blob/33f49f33d6f8cec9a5772bc7f3a4e5f1de207b30/clippy_lints/src/enum_variants.rs#L107-L112).

marker_lints/src/not_using_has_span_trait.rs
Veetaha: I've noticed that there is no information about the parentheses around the type in `TyKind`s. For example both `u32`, `(u32)`, and `((u32))` are represented the same in the AST. This is in contrast to `syn`, where it has a [`Paren`](https://docs.rs/syn/latest/syn/struct.TypeParen.html) type kind that wraps the inner type in parens explicitly.

I added two test cases to `print_ty.rs` to show that there is no info about that. Is this by design? I don't think this is critical, and in fact the `Paren` type kind in `syn` is often annoying and I have to write the same `strip_ty_group` function that removes the parentheses and an invisible [`Group`](https://docs.rs/syn/latest/syn/struct.TypeGroup.html). It's really hard to understand what `Group` is, but I suppose that's related to invisible groupping that is added by the compiler when a macro is invoked in type position. For example `&dyn macro_call!()` is implicitly grouped when expanded like `&dyn (expanded_ty_here)` and the parentheses I used here aren't visible.

marker_lints/src/not_using_has_span_trait.rs
xFrednet: Clippy usually has one lint pass per lint crate, but we also combine them sometimes. This can become messy (See [`clippy_lints::methods`](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/methods/mod.rs) as an extreme example), but allows lints to share common logic.

We could do something similar, where we have a lint pass trait implemented on several types. The main `LintPass` of the crate would hold `Vec<Box<&dyn LintPass>>`, to distribute the control flow inside the crate. The `dyn LintPass` wouldn't have access to the `'ast` lifetime, but I think we want to discourage users from storing any `'ast` objects in the structs anyways

marker_lints/src/not_using_has_span_trait.rs
xFrednet: > Is this by design?

Yes, such information is usually not represented in ASTs, as they are semantically the same. You'll see the same thing for binary operators. `(1 + 2) * 3` the brackets are not included in the final AST. The tree structure preserves the execution order.

This is part of what the abstract from "Abstract Syntax Tree" is about. It tries to unify these syntactical differences into one uniform view, at the loss of information. If a user wants, they can still use the span, to see if there are parentheses, but that's not necessarily recommended.

Clippy has a few lints, which check for the syntax. But they're always a bit weird, as we use Clippy as a formatter with semantic information. [clippy::semicolon_if_nothing_returned](https://rust-lang.github.io/rust-clippy/master/index.html#/semicolon_if_nothing_returned) is an example, where we need to check the written syntax as well for some heuristics, if the semicolon should be suggested.

> So I'm not really sure marker needs to expose any parens info at all. Stripping them automatically simplifies life, but maybe there could be lints that want that info? I'm not sure.

In Clippy we have lived happy without this information, and hacked around with spans, where needed. If we wanted to provide a way to completely reconstruct the original source code, we should include parenthesis. Rustfmt and syn, for instance, require this information. For Marker's case, I don't believe that this should be one of our goals. It would be nice to have a function to print the AST as it would be written (Similar how rustc has this for hir, with the "Show HIR" option on the playground). But adding parenthesis nodes, would add a lot of edge cases to linting, with very little benefit :)

marker_lints/src/not_using_has_span_trait.rs
Veetaha: Aren't clippy lints all in one crate? I've shown the example where they register one lint pass per lint module.

That extreme example is really extreme. It's hard to tell which logic they share in common, and wether it's even worth it. I suppose it's an attempt to save some performance on looking up method expressions in code. I.e. the difference between:
```rust
if let Expr::Method(method) = expr {
	method_lint_1(method);
	method_lint_2(method);
}
```
```rust
if let Expr::Method(method) = expr {
	method_lint_1(method);
}

if let Expr::Method(method) = expr {
	method_lint_2(method);
}
```
and the first one wins because it doesn't have a repeated check?
That makes sense, but I'm not sure such optimization is worth it, given the messy code organization, but I suppose we shouldn't have an API that disallows doing such context sharing.

Maybe there could be several variations of `declare_lint!` macro. One for the simple case where the lint with the lint pass implementation is directly registered in `inventory`, and the second one more advanced that registers only the lint in the inventory, but the implementation of the lint pass is expected to be used manually inside of other lint pass implementation and thus not registered.

marker_lints/src/not_using_has_span_trait.rs
xFrednet: Rustc has a util to convert AST types to semantic types, it can crash under some circumstances, which is why I'm cautious with it. I think the better option would be to have a `resolve_item_ids` equivalent. The AST path of the `ast::TyKind` should already provide the `ItemId` it refers to. Right now, there is just no way to check if that's the one from our `Span` struct

xFrednet (approved):
Looks good to me, thank you for the nice update and new lint :heart:

Issue #236, opened 2023-09-06
Bump Nightly 2023-10-05


xFrednet:
Rust PRs which will probably influence the next nightly bump

* https://github.com/rust-lang/rust/pull/115507
* https://github.com/rust-lang/rust/pull/115418
* https://github.com/rust-lang/rust/pull/116163

PR #281, opened 2023-10-08
Chore: Bump nightly -> `2023-08-24`

xFrednet:
Closes: https://github.com/rust-marker/marker/issues/236

### DIFF
--- a/scripts/release/install.ps1
+++ b/scripts/release/install.ps1
@@ -49,7 +49,7 @@ $ErrorActionPreference = "Stop"
 $version = "0.3.0"
 # endregion replace-version unstable
 
-$toolchain = "nightly-2023-08-24"
+$toolchain = "nightly-2023-10-05"
 
 # Log the command, execute, and fail if its exit code is non-zero.
 # Surprisingly PowerShell can't do the exit code checks for us out of the box.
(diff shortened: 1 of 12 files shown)


scripts/release/install.ps1
```diff
@@ -49,7 +49,7 @@ $ErrorActionPreference = "Stop"
 $version = "0.3.0"
 # endregion replace-version unstable
 
-$toolchain = "nightly-2023-08-24"
+$toolchain = "nightly-2023-10-05"
```
xFrednet, line 52: Looking at this and the documentation update changes, I think, it would be better to switch the nightly toolchain script to also use region markers. This would also allow us to have different variations. This install script still says version v0.3.0, but now has a newer toolchain, which is not ideal...

Veetaha: Vector uses unstable trait specialization feature, so it should definitely be optimized for the case when the `Vec` is extend with another `Vec` ([code](https://github.com/rust-lang/rust/blob/4f4a413fe6931d0ad9d3ac6bd20ff36398961e64/library/alloc/src/vec/mod.rs#L2823)).

xFrednet:
Draft: blocked on https://github.com/rust-marker/marker/pull/284

PR #282, opened 2023-10-08
Print use items, add more docs

Veetaha:


marker_api/src/ast/item/use_item.rs
```diff
@@ -19,6 +19,11 @@ use super::CommonItemData;
 /// // `use_kind()` -> `Single`
 /// ```
 ///
+/// There is a separate [`UseItem`] per each path in a `use` declaration.
+/// For example, if you have `use foo::{bar, baz};` there will be two [`UseItem`]s
+/// one for `bar` and one for `baz` and they will share a prefix in their
+/// [`UseItem::use_path()`].
```
xFrednet, line 25: Nice catch. I thought, I had already documented this behavior, but apparently not. Before 1.0 we might want to revisit this representation, but for now, this is a solid improvement :D

xFrednet (approved):
LGTM

How do you usually like to handle PR merges? I've seen you approve PRs, but leave the merge action to me. Now that we can both merge, do you have a preference? Either way, you can merge this PR :)

Veetaha:
> How do you usually like to handle PR merges?

I don't have strict rules. If there are no additional comments from you, you can merge it yourself.
If you approve, but there are some minor changes suggested, I can fix them without your re-review and merge once commits are pushed.

The problem is with squashing the commits though. I prefer having a single commit per PR, but GitHub's merge queues have a shitty squash-and-merge config that just concatenates all commit messages into one. I guess we could live with that, because the main thing that I care in the commit is that it must have a clickable reference to the PR where it came from in its title.

xFrednet:
I thought about switching over to squash and merge commits. I didn't want that for the monster PR's I created with 500+ line changes (totally speaking in past tense, here xD). Now it could be something worth discussing again :)

Veetaha:
THe size of the PR doesn't matter to me. I merged PRs with 5K LoC changes in one commit, especially because a large PR often has a dirty commit history of initial attempts, then rewrites, then new attempts and new rewrites, then merge conflicts resolving, updating to the new master and stuff which is just noise that one wouldn't like to see in master.

Issue #251, opened 2023-09-21
[Feat]: Introduce a lock file for lint crates


Veetaha:
### Summary

# Context
Today `Cargo.lock` file for the internal cargo workspace created for building the lint crates isn't exposed to the users of marker.
This means that every time `cargo marker` would run on CI the latest compatible versions of lint crates and their dependencies will be built.

Even though, ideally one would think it should not cause any problem, but the real world is cruel. People may release a breaking change by mistake, or some crate could rely on the behaviour not included in its dependency semver guarantees. For example, I remember some crates that don't bump their major version if they increase the MSRV, or there are crates that simply don't care about the MSRV. There are so many reasons one could fuck up the semver compatibility..

It's also a security concern. You don't want to always use the latest version of your dependencies. That opens the possibility for any of the dependencies to introduce a malicious build script in the new patch version that would leak CI credentials.

# Summary

This all led to the creation of `Cargo.lock` concept. Marker needs to continue exposing that to the users for its lint crates.

I don't have a final design for this yet. What I see right now is the following.

Marker will generate a `Marker.lock` which is going to be just the copy of the `Cargo.lock` that marker uses internally. Marker will copy that lock file into the internal cargo workspace each time it does a build of the lint crates.

We need to provide the tools for updating the lock file `cargo update`, generating it `cargo generate-lockfile`, reviewing the dependency tree `cargo tree`, auditing it [`cargo deny`](https://github.com/EmbarkStudios/cargo-deny), etc. 

Basically, we need to expose the users to managing the lock file for the internal marker's workspace.
I wouldn't like to have to write all the possible tools for managing the lock file in `cargo-marker` by hand. Besides the 3-rd party cargo plugins like `cargo-deny` or `cargo-audit` and any other ones that may arise in the future should also be supported.

So I think there should be a dynamic `cargo` command wrapper in `cargo-marker` to run them inside the internal marker cargo workspace and update the `Marker.lock`. For example a new `cargo` command that allows running `cargo` inside of the internal workspace and it just forwards whatever is passed after `cargo marker cargo` to `cargo` inside of that workspace.
```
cargo marker cargo update
cargo marker cargo tree
cargo marker cargo deny
```
Naming could probably be improved here, but this is my first thought

xFrednet:
Big thumps up, this is something I haven't considered thus far!

I like the idea of providing a dynamic cargo command wrapper. For the naming, we could use something like `lints`, `lint-crates`, `cargo-lints`?

[...]

PR #283, opened 2023-10-08
Remove the need to use `register_tool`, and add it automatically.

Veetaha:
This is a bit of a hack, because `rustc`'s code wasn't designed for such a use case, so the code looks a bit ugly, but it works.

The solution is to add a query override for `registered_tools` that adds marker automatically.

I updated the book, but I figured out that the book needs to be updated only after the release. I wonder if we should make the book deployment on release workflow and not on each push to master?


marker_rustc_driver/src/main.rs
```diff
@@ -75,6 +76,35 @@ impl rustc_driver::Callbacks for MarkerCallback {
             register_tracked_files(sess, &lint_crates);
         }));
 
+        // Make it possible to use `#[allow(marker::{lint_name})]` without
+        // having to add `#![feature(register_tool)]` and `#![register_tool(marker)]`.
+        config.override_queries = Some(|_sess, providers, _extern_providers| {
```
xFrednet, line 81: I didn't even know that it was possible to override quarries o.O. 

In red pen, I found another hack, that adds the attributes via console arguments. Have you considered that as well?

https://github.com/estebank/redpen/blob/b4008d2092ff4fdbb3f735d1892a88a63c956d97/commands/wrapper.rs#L39-L40

I'm not sure if that's more or less sustainable than this option.

The best would be to have an official interface in rustc for this.

xFrednet:
>  I wonder if we should make the book deployment on release workflow and not on each push to master?

That might be good and would avoid confusion in the future. :+1:

marker_rustc_driver/src/main.rs
Veetaha: Yeah, that may be a good idea. I wonder if it could be configured at workspace level, if that's a cli option it could be part of rustflags, I'll think what it may give us

Veetaha: I though that we could have a thin rustc wrapper script e.g. `cargo marker rustc` that users could put into `RUSTC_WRAPPER` to avoid writing `cfg_attr` and any other ugly boilerplate. That wrapper would add the `-Zcrate-attr=feature(register_tool)`, and `-Zcrate-attr=register_tool(marker)` automatically to any `rustc` invocation. But.. there is a problem with this approach:

`-Zcrate-attr` is an unstable option, so users can't have it on stable toolchain. We could potentially add `RUSTC_BOOTSTRAP=1` env variable to workaround that, but that env variable affects the entire `rustc` CLI invocation, meaning users can now use any unstable options, because our wrapper script adds that env variable for them. That's quite ugly..

So for having this inside of the marker rustc driver only and requiring `cfg_attr` is the best option for now.

Veetaha:
I changed the approach to use the CLI options. I actually supposed there could be an option like this, but I failed to find it in rustc docs and help messages. The problem is that I didn't run `rustc -Z help` to see the help message for the unstable options.
But luckily, you pointed out that it exists, and it looks much better than a crutch query override with a thread-local.

docs/book/src/usage/setting-lint-levels.md
```diff
@@ -3,52 +3,20 @@ Once your crate is configured to run some lints, it quickly becomes important to
 code. Marker provides the ability to use normal lint control attributes like `#[allow(...)]` `#[deny(...)]` and others
 to control the behavior of marker lints.
 
-## On nightly
-Marker uses the `marker::` tool prefix for lints. In order for rustc to recognize this prefix, it requires marker to be
-registered via the [`#[register_tool()]` feature](https://github.com/rust-lang/rust/issues/66079).
+Marker uses the `marker::` tool prefix for lints. You can then use all of the normal lint attributes you might usually use, on marker provided lints, like `#[allow(marker::my_lint)]`, but you have to put them under `#[cfg_attr(marker, ...)]`. This is because normally `rustc` doesn't know anything about `marker` and it will complain about an unknown lint if yo u run simple `cargo check/build`.
```
xFrednet, line 6:
```suggestion
Marker uses the `marker::` tool prefix for lints. You can then use all of the normal lint attributes you might usually use, on marker provided lints, like `#[allow(marker::my_lint)]`, but you have to put them under `#[cfg_attr(marker, ...)]`. This is because normally `rustc` doesn't know anything about `marker` and it will complain about an unknown lint if you run simple `cargo check/build`.
```

docs/book/src/usage/setting-lint-levels.md
```diff
@@ -3,52 +3,20 @@ Once your crate is configured to run some lints, it quickly becomes important to
...
+Marker provides config arguments to `rustc` during the lint passes, so that your linted code can conditionally apply attributes only during the marker run.
 
 Specifically marker passes `--cfg=marker` and a `--cfg=marker="my_crate"` flag for each lint crate. This means that you
 can use `#[cfg_attr(marker, foo)]` to conditionally apply the `foo` attribute only during lint runs.
 
-This conditional compilation can be used to leverage the fact that marker uses nightly for linting, without requiring
-the project to use a nightly toolchain. To do this, add `#![cfg_attr(marker, feature(register_tool))]` and
-`#![cfg_attr(marker, register_tool(marker))]` attributes to the top of your crate file, to register marker as a tool.
-Then you can apply lint level attributes like `#[cfg_attr(marker, allow(marker::foo))]` to control your marker lints.
+This conditional compilation can be used to leverage the fact that marker uses a custom rustc implementation, without requiring the project to use that implementation by default. Then you can apply lint level attributes like `#[cfg_attr(marker, allow(marker::foo))]` to control your marker lints.
```
xFrednet, line 13: This paragraph confuses me. What's the specific thing which you want to say? It seems like the `--cfg marker` was already mentioned and explained

xFrednet (commented):
The implementation looks good to me, but parts of the documentation are confusing to me :sweat_smile:

xFrednet:
> I wonder if we should make the book deployment on release workflow and not on each push to master?

I've crated #285 to only deploy the book on tags or when it was manually triggered :)

Veetaha:
I've been trying to preserve the previous sentences from the documentation, and as you consequently noticed it made it repetitive and confusing.

Instead, I rewrote it and also included the requirement of adding the crate name into the lint name which is going to be part of https://github.com/rust-marker/marker/pull/288

xFrednet:
> WDYT @xFrednet. How do you feel about removing `--cfg=marker="lint_crate"`?

I feel rather strongly about keeping it. This cfg is targeted towards lint crates, which can be time intensive, like architecture checks. In those cases, users would probably not want to run them on their local machine before committing, but let the CI handle it. With this feature, it's possible to only apply the attribute if the lint crate is actually loaded.

I can also imagine that there could be lints for specific circumstances, like releases. Clippy has a few settings/lints which are allow-by-default, but can be worth checking before a release. This could also be a thing in Marker, with the caveat, that these lints/lint crates could be in a separate lint crate to safe performance.

If we don't provide a way to check for lint crate, it would mean that users can't set those lint levels at all or that they have to manually add a feature to check modify the lint levels again.

Veetaha:
I think heavy lints will be "allow" by default, and then on CI their level will be overridden with `RUSTFLAGS = --warn {lint_name}`.
Could you show a specific example which one couldn't solve nicely without having the `--cfg=marker="{lint_crate}"`?

To my mind people will have a single `cargo marker` CI job with a bunch of heavy lints enabled via `RUSTFLAGS`. Locally they will be ignored because they are "allow" by default. So there will never be a case of "unknown lint" warning.

xFrednet:
The lint level doesn't prevent the lint logic from being run. Rustc still calls every lint pass, regardless of if the lint is currently allowed or not. There is some discussion going on, if this can be optimized, but there are several edge cases which make this difficult. For example, a lint could be emitted at a different node than the one that is currently being checked. Marker has it especially difficult, since we currently force all lints to be implemented in the same lint pass.

Users can of course, always include a check, if the lint is currently allowed, but that adds more burden on them. Then there are also lint expectations, using the `#[expect]` attribute, which require the lint emission, even if it's not shown to the user.

I assume that architecture lints also first need to collect a lot of information, to then do one final analysis at the end. Meaning that not running any logic in them, if they are allowed, might be unexpected behavior.

All this is to say, that having a lint be allowed, still makes it run and doesn't save the CPU/Memory usage.

---

I once worked at a company, where we had some architecture lints using [ArchUnit](https://www.archunit.org/userguide/html/000_Index.html#_asserting_architectural_constraints) for Java. I would like to have something similar built on top of Marker later down the line. But such checks require building an entire call graph, dependency graph... Maybe even across crate boundaries.

This is a case, where it's better to only run the lints in the CI. The lints have a low hit rate, since they're quite specific or only focus on a subset of structs. Running them on the user dev machines, will result in nothing in 99% of the cases. Running them on CI if any `.rs` files have been changed will still have a low hit rate, but won't slow down the end users.

So this would be an example, where I would only load the lint crate in the CI. Does this example make sense? I'm happy to give some more details or provide other examples :)

---

We might be able to add an option to Marker, to only load the lint names, from a lint crate without actually running them. But IDK, if that's the best solution and even then, if that would cover every use case.

Veetaha:
I added the docs `--cfg=marker="lint_crate"`. Although, while writing them and thinking about this I found a potential problem with the `--lints` CLI parameter in light of #251. If we allow specifying lints in the CLI then how do these lints get into the `Marker.lock`? Maybe there won't be a problem here if we use the lock file of the lint itself and not have a `Marker.lock`. That is a good argument against `Marker.lock`

Veetaha:
I rebased after https://github.com/rust-marker/marker/pull/288. You may check the changes after the commit [Add docs about --cfg=marker="lint_crate"](https://github.com/rust-marker/marker/pull/283/commits/44d021fc4efe2600c59773da74129e0bb17e906a)

xFrednet:
With the current behavior of `--lints`, which only loads the lint crate specified via console args, it can cause the `unknown_lints` lint to trigger from rustc. I wish there was a way to easily intercept that lint from our side, but it doesn't seem that way, from the rustc code I checked. If Marker gains more traction, we should be able to add some special handling to rustc for it. A hacky solution we could do, is just allow the lint, when running marker, and have our own `unknown_lints` version which is lint crate aware. Then we could remove the `--cfg=marker="{lint_crate}"` again :thinking: This might fail if users have a `#[warn(unknown_lints)]` attribute in their code, but I've never seen that, so it should be fine.

I actually like this idea a lot. One question would be, where we store buildin lints (Maybe a lint crate or `marker_adapter`?) but that should not be a blocker.

Do you want to remove the added section in favor of this solution or wait for the implementation instead? Note that such a lint implementation is currently blocked, since attributes are not supported yet.

---

For `Marker.lock`, your right that this could be a problem. I have two ideas, which both feel hacky, but they might be a good inspiration:
1. Compile each lint crate individually. And combine them all somehow, for example in a directory or by concatenating the `Cargo.lock` files in a `Marker.lock` file. This would support command line lints as well. However, it would require us to manually separate the lock file part for every compilation and will make it difficult to run simple cargo comments on the dummy fetch crate.
2. Store the `Marker.lock` file for command line lints separately, maybe under a global `~/.marker` folder?

xFrednet (approved):
LGTM, I think we should have the last paragraph about the lint logic being run regardless of the lint level in this documentation, even if we decide to use a custom `unknown_lints` lint for Marker instead :)

xFrednet:
I've moved our `Marker.lock` comments to #251 to have them all in one place :)

Veetaha:
> Do you want to remove the added section in favor of this solution, or wait for the implementation instead? Note that such a lint implementation is currently blocked, since attributes are not supported yet.

Yeah the current `--cfg=marker="{lint_crate}"` looks a bit ugly, but at least not magical. I would like to keep it docs at least because  I just want to get this PR merged 😅. We can decide and change how it works separately. So I'm going to merge it.

---

As for the problem of triggering `unknown_lints`. Overriding the default `unknown_lints` lint and providing a custom impl seems like it would be too much of a hack. Besides... what will this custom impl do? Will it never report an unknown lint warning? It is still quite a nice feature to be warned if you made a typo in the lint name.

Maybe a lesser hack could be having a specialized feature for lint "profiles". This is similar to cargo build profiles or [`cargo-nextest` profiles](https://nexte.st/book/configuration.html).

Users could define several profiles that would extend the default list of lints. They will then be able to select additional lints configured in `Cargo.toml` metadata using a `--profile {profile_name}` argument. This solution will then deprecate the `--lints` CLI parameter to avoid problems with `Marker.lock` as well.

Having everything specified in a config file allows Marker to know which lint crates the project may ever expect to be run with. So when marker is run with the default profile that doesn't include heavy lints Marker will still know that there are other lint crates that could participate in linting, and it could provide noop lint implementations for lints from those crates.

One problem with this is that Marker would need to compile all lint crates specified in the config including lint crates from other profiles. Maybe if lint crates also had some metadata that could list their lints without compiling them this could solve this problem.

xFrednet:
> As for the problem of triggering unknown_lints. Overriding the default unknown_lints lint and providing a custom impl seems like it would be too much of a hack. Besides... what will this custom impl do? Will it never report an unknown lint warning? It is still quite a nice feature to be warned if you made a typo in the lint name.

The lint would check lint level attributes. With #288 we always know where a lint comes from just by looking at the identifier. The lint would basically do the following:
1. Ignore any lint, which doesn't have the `marker::` prefix. Rustc can handle that during normal compilation.
2. If it encounters a lint with the `marker::` prefix, it checks if the specified lint crate is loaded.
    * If it's not loaded everything is fine, and we just accept the name, without any comment, since it might be in the lint crate
    * If the lint crate is loaded, we check if the name is included in the list of registered lints, we got from the lint crate during startup. If it's included, awesome. Otherwise, we emit a warning with the `marker::unknown_lint` lint.

---

Having lint profiles sounds like a good and flexible idea :+1:

PR #285, opened 2023-10-10
CI: Only deplay GH Pages on release

xFrednet:
This delays the deployment of our book until a version tag is pushed, or the workflow is manually triggered. I don't expect us to manually trigger it often, but it's nice to have the option.

I also wanted to swap two chapters, so I might manually trigger the deployment after this PR ^^

---

I've considered having a model like Clippy has for it's GH Pages, where we store a snapshot for every version. But I don't think this is really useful for us rn. Having a single deployment is simpler has a cleaner URL.

It also means that there is only one version and users can't get confused. In Clippy, we had a few people, that were confused why the GH pages of the "master" version weren't inline with their experience on the stable toolchain.

---

Prerequisite for rust-marker/marker#283


Veetaha (approved):
LGTM, although I have one small fix to include in the current book:
https://github.com/rust-marker/marker/pull/287
Please merge it to fix the released book.

I believe the `workflow_dispatch` is actually a good way to backport fixes into the released book. This way, for example, if we fixed a broken link, you could `git cherry-pick` the commit with the fix into a custom branch named `hotfix-0.3`, which is based on `0.3.0` git tag. We could push any hotfixes of the book there, run the book build from that branch, and then remove the branch once `0.4.0` is released with the new version of the book now available officially with all the fixes already in it.

Or we can leave that `hotfix-0.3` branch for the history, because we could use that branch not only for the book hotfixes, but also for making new patch versions in old releases. E.g. if `0.4.0` is already released, but there was some security bug in `0.3.0` that we need to fix in the old version we could push it into a `hotfix-0.3` branch and then run the release workflow from that branch. The release workflow was created with the assumption that it could be run from a non-master branch.

PR #288, opened 2023-10-11
API: Include the lint crate in the lint identifier

xFrednet:
This PR adds the lint crate name, as an infix for the lint identifier.

* `marker::diag_msg_uppercase_start` -> `marker::marker_lints::diag_msg_uppercase_start`

This prevents lint name collisions across different lint crates.

---

This is a major breaking change, but I think it'll definitely be worth it in the long run. We might want to add a way to rename lint crates, this would allow users to shorten some names, like this:

* `marker::super_lints::diag_msg_uppercase_start` -> `marker::super::diag_msg_uppercase_start`

---

Inspired by rust-marker/marker#278


.github/workflows/gh-pages.yml
```diff
@@ -10,7 +10,7 @@ on:
     # Match only on specific tags. We don't want this workflow to be invoked when
     # we put sliding `v{major}` and `v{major}.{minor}` tags on the same commit.
     tags: ['v[0-9]+.[0-9]+.[0-9]+*']
-    workflow_dispatch:
+  workflow_dispatch:
```
Veetaha, line 13: This seems like an accidental change

CHANGELOG.md
```diff
@@ -27,6 +27,11 @@ The [v0.4.0 milestone] contains a list of planned changes.
 
 [v0.4.0 milestone]: https://github.com/rust-marker/marker/milestone/4
 [#278]: https://github.com/rust-marker/marker/pull/278
+[#288]: https://github.com/rust-marker/marker/pull/288
+
+### Changes
+
+- [288]: Lint identifiers use the lint crate name as a new infix, to prevent name collisions across lint crates.
```
Veetaha, line 34: This is a breaking change, and thus needs to be added in the `Breaking Changes` section below instead.

marker_lints/tests/uitest.rs
```diff
@@ -2,10 +2,15 @@ use marker_uitest::ui_test::*;
...
 fn main() -> color_eyre::Result<()> {
-    let mut config = marker_uitest::simple_ui_test_config!("tests/ui", "../target")?;
+    let mut config: Config = marker_uitest::simple_ui_test_config!("tests/ui", "../target")?;
 
     config.dependencies_crate_manifest_path = Some("./Cargo.toml".into());
 
+    // `marker_api::declare_lint` uses the `CARGO_CRATE_NAME` environment value.
+    // Setting it here once, makes sure that it'll be available, during the
+    // compilation of all ui test files.
+    std::env::set_var("CARGO_CRATE_NAME", "marker_lints_uitests");
```
Veetaha, line 12: This looks like a painful papercut. I don't understand how it works. Why is this `marker_lints_uitests` and not just `marker_lints`? The name `marker_lints_uitests` isn't mentioned anywhere else in this PR. I also don't see this code setup for `marker_uilints`.

Could this be avoided? Or could the workaround be encapsulated inside of `marker_uitest::simple_ui_test_config`?

Veetaha (commented):
Beware that I changed the book chapter about specifying lints in https://github.com/rust-marker/marker/pull/283. The docs there already include the requirement of specifying the lint crate name in the lint name.

.github/workflows/gh-pages.yml
xFrednet: Nope, I should have left a comment on the PR. The `workflow_dispatch` item is currently under the `push` action, which means that it can't be triggered manually. This moves it under the `on:` item

marker_lints/tests/uitest.rs
xFrednet: `ui_test` doesn't use Cargo and instead calls the rustc driver directly. This is for a few reasons, as Cargo would need a workspace, `Cargo.toml` file and would also be a bit slower for only compiling a single file. However, this also means that any environment values usually set by Cargo (starting with `CARGO_*`) are not set when the uitest files are compiled. 

This should really only be noticed, if someone writes lints for the `marker_api::declare_lint!` macro. That's why only the uitest of `marker_lints` required this environment value to be set. So I don't really see it as a paper cut.

.github/workflows/gh-pages.yml
Veetaha: Ah, I see. I suppose github would show a warning if this workflow is run

xFrednet: Imagine having a tool to statically analyze source code or files. That would be a cool invention :wink: xD

xFrednet:
I assume this PR is ready, since all comments have been addressed. Merging this will also unblock the docs from #283 

Thank you for the review and documentation update in #283

PR #290, opened 2023-10-12
Don't publish the book on pre-releases

Veetaha:
Right now doing a pre-release is a bit broken due to the problem with the toolchain update. I don't know if we even really need pre-releases and how soon, but I don't think anything prevents us from having the ability to do these.

This is a small fix for the book deployment which is just one of the things affected by pre-releases. It makes sense not to deploy the book during a pre-release because it will contain the docs for the new release and not for the stable one.


xFrednet (approved):

xFrednet:
Good point, the `-*` suffix was some good old copy-pasta :spaghetti:

PR #294, opened 2023-10-18
API: Make item `Visibility` visible in the API

xFrednet:
Designing a sensible `Visibility` representation is surprisingly hard. I had to update the function names several times, and still feel like they can be improved. There is also still one util function missing, which requires the implementation of https://github.com/rust-marker/marker/issues/242 first. I've left a corresponding comment in the PR.

I think this implementation will cover 99% of the current use cases. This PR also updates the `not_using_has_span_trait` lint and enables it by default inside the `marker_api` crate

---

Closes: rust-marker/marker#26

### DIFF
--- a/marker_api/src/ast/item.rs
+++ b/marker_api/src/ast/item.rs
@@ -190,6 +190,7 @@ use impl_item_type_fn;
 #[repr(C)]
 #[derive(Debug)]
 #[cfg_attr(feature = "driver-api", visibility::make(pub))]
+#[cfg_attr(feature = "driver-api", derive(typed_builder::TypedBuilder))]
 struct CommonItemData<'ast> {
     id: ItemId,
     span: SpanId,
@@ -238,15 +239,6 @@ macro_rules! impl_item_data {
 
 use impl_item_data;
 
-#[cfg(feature = "driver-api")]
-impl<'ast> CommonItemData<'ast> {
-    // Here we can't `derive(TypedBuilder)`, as the created builder is private,
-    // due to the declaration of the item not being pub.
-    pub fn new(id: ItemId, span: SpanId, vis: Visibility<'ast>, ident: Ident<'ast>) -> Self {
-        Self { id, span, vis, ident }
-    }
-}
-
 /// The declared visibility of an item or field.
 ///
 /// Note that this is only the syntactic visibility. The item or field might be
@@ -274,15 +266,9 @@ impl<'ast> CommonItemData<'ast> {
 pub struct Visibility<'ast> {
     #[cfg_attr(feature = "driver-api", builder(setter(skip), default))]
     _lifetime: PhantomData<&'ast ()>,
-    #[cfg_attr(feature = "driver-api", builder(setter(into)))]
-    #[cfg_attr(feature = "driver-api", builder(default = FfiOption::None))]
+    #[cfg_attr(feature = "driver-api", builder(setter(into), default))]
     span: FfiOption<SpanId>,
     kind: VisibilityKind,
-    // * The rustc model makes sense, so have a kind with a restriction
-    // * have according methods, that check for `pub(crate)`
-    // * document possible equivalency between `pub(crate)` and `pub(super)`
-    // * Document that things can be reexported
-    // * Test how visibility only to lower scopes works `pub(self::sub_module)`
 }
 
 impl<'ast> Visibility<'ast> {
@@ -321,8 +307,8 @@ impl<'ast> Visibility<'ast> {
     /// This function checks if the visibility is restricted and the defined path
     /// belongs to the root module of the crate. Meaning, that this can also be `true`,
     /// if the visibility uses `pub(super)` to travel up to the crate root.
-    ///
-    /// ```rs
+    // Ignore, since the `in crate::example_1` path doesn't work for doc tests
+    /// ```ignore
     /// // lib.rs
     ///
     /// mod example_1 {
@@ -333,11 +319,11 @@ impl<'ast> Visibility<'ast> {
     ///     pub(in crate::example_1) fn bar() {}
     ///
     ///     // Returns `true` as the visibility is restricted to the root of the crate.
-    ///     pub(crate) baz() {}
+    ///     pub(crate) fn baz() {}
     ///
     ///     // Returns `true` as the visibility is restricted to `super` which is the
     ///     // root of the crate.
-    ///     pub(crate) boo() {}
+    ///     pub(super) fn boo() {}
     /// }
     ///
     /// // Returns `false` since the visibility is not restricted
@@ -453,14 +439,10 @@ impl<'ast> Body<'ast> {
 
 #[cfg(all(test, target_arch = "x86_64", target_pointer_width = "64"))]
 mod test {
-    use super::*;
-    use expect_test::{expect, Expect};
+    use crate::test::assert_size_of;
 
-    #[track_caller]
-    fn assert_size_of<T>(expected: &Expect) {
-        let actual = std::mem::size_of::<T>();
-        expected.assert_eq(&actual.to_string());
-    }
+    use super::*;
+    use expect_test::expect;
 
     #[test]
     fn test_item_struct_size() {
--- a/marker_api/src/ast/item.rs
+++ b/marker_api/src/ast/item.rs
@@ -333,8 +333,8 @@ impl<'ast> Visibility<'ast> {
         matches!(self.kind, VisibilityKind::Crate(_))
     }
 
-    /// Returns `true` if a visibility has been defined.
-    pub fn is_defined(&self) -> bool {
+    /// Returns `true` if a visibility has been explicitly specified.
+    pub fn is_explicit(&self) -> bool {
         !matches!(self.kind, VisibilityKind::Default(_) | VisibilityKind::DefaultPub)
     }
 
--- a/marker_api/src/ast/item.rs
+++ b/marker_api/src/ast/item.rs
@@ -454,25 +454,31 @@ impl<'ast> Body<'ast> {
 #[cfg(all(test, target_arch = "x86_64", target_pointer_width = "64"))]
 mod test {
     use super::*;
-    use std::mem::size_of;
+    use expect_test::{expect, Expect};
+
+    #[track_caller]
+    fn assert_size_of<T>(expected: &Expect) {
+        let actual = std::mem::size_of::<T>();
+        expected.assert_eq(&actual.to_string());
+    }
 
     #[test]
     fn test_item_struct_size() {
         // These sizes are allowed to change, this is just a check to have a
         // general overview and to prevent accidental changes
-        assert_eq!(56, size_of::<ModItem<'_>>(), "ModItem");
-        assert_eq!(48, size_of::<ExternCrateItem<'_>>(), "ExternCrateItem");
-        assert_eq!(64, size_of::<UseItem<'_>>(), "UseItem");
-        assert_eq!(80, size_of::<StaticItem<'_>>(), "StaticItem");
-        assert_eq!(72, size_of::<ConstItem<'_>>(), "ConstItem");
-        assert_eq!(144, size_of::<FnItem<'_>>(), "FnItem");
-        assert_eq!(112, size_of::<TyAliasItem<'_>>(), "TyAliasItem");
-        assert_eq!(96, size_of::<StructItem<'_>>(), "StructItem");
-        assert_eq!(88, size_of::<EnumItem<'_>>(), "EnumItem");
-        assert_eq!(88, size_of::<UnionItem<'_>>(), "UnionItem");
-        assert_eq!(112, size_of::<TraitItem<'_>>(), "TraitItem");
-        assert_eq!(144, size_of::<ImplItem<'_>>(), "ImplItem");
-        assert_eq!(64, size_of::<ExternBlockItem<'_>>(), "ExternBlockItem");
-        assert_eq!(48, size_of::<UnstableItem<'_>>(), "UnstableItem");
+        assert_size_of::<ModItem<'_>>(&expect!["80"]);
+        assert_size_of::<ExternCrateItem<'_>>(&expect!["72"]);
+        assert_size_of::<UseItem<'_>>(&expect!["88"]);
+        assert_size_of::<StaticItem<'_>>(&expect!["104"]);
+        assert_size_of::<ConstItem<'_>>(&expect!["96"]);
+        assert_size_of::<FnItem<'_>>(&expect!["168"]);
+        assert_size_of::<TyAliasItem<'_>>(&expect!["136"]);
+        assert_size_of::<StructItem<'_>>(&expect!["120"]);
+        assert_size_of::<EnumItem<'_>>(&expect!["112"]);
+        assert_size_of::<UnionItem<'_>>(&expect!["112"]);
+        assert_size_of::<TraitItem<'_>>(&expect!["136"]);
+        assert_size_of::<ImplItem<'_>>(&expect!["168"]);
+        assert_size_of::<ExternBlockItem<'_>>(&expect!["88"]);
+        assert_size_of::<UnstableItem<'_>>(&expect!["72"]);
     }
 }
--- a/marker_api/src/ast/item.rs
+++ b/marker_api/src/ast/item.rs
@@ -1,4 +1,4 @@
-use std::{fmt::Debug, marker::PhantomData};
+use std::fmt::Debug;
 
 use crate::{
     common::{HasNodeId, ItemId, SpanId},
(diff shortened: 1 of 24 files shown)


marker_api/src/ast/item.rs
```diff
@@ -238,42 +240,182 @@ use impl_item_data;
...
-    pub fn new(id: ItemId, span: SpanId, ident: Ident<'ast>) -> Self {
-        Self {
-            id,
-            span,
-            vis: Visibility::new(id),
-            ident,
-        }
+    // Here we can't `derive(TypedBuilder)`, as the created builder is private,
+    // due to the declaration of the item not being pub.
+    pub fn new(id: ItemId, span: SpanId, vis: Visibility<'ast>, ident: Ident<'ast>) -> Self {
```
Veetaha, line 245: This surprised me, so I checked and the following works. You just need to make sure that `visibility::make(pub)` preceds the `TypedBuilder` derive due to the order of macro expansion.
```rust
#[repr(C)]
#[derive(Debug)]
#[cfg_attr(feature = "driver-api", visibility::make(pub))]
#[cfg_attr(feature = "driver-api", derive(typed_builder::TypedBuilder))]
struct CommonItemData<'ast> {
    id: ItemId,
    span: SpanId,
    vis: Visibility<'ast>,
    ident: Ident<'ast>,
}
```

Here, for example, it already works like that.
https://github.com/rust-marker/marker/blob/73835de0a82ebd33b959df200f66e5a3d7bf007a/marker_api/src/ast/stmt.rs#L59-L73

Veetaha (commented):
I'll take a closer look on weekends, didn't review everything yet.

marker_api/src/ast/item.rs
```diff
@@ -238,42 +240,182 @@ use impl_item_data;
...
+    pub fn is_pub_with_path(&self) -> bool {
+        matches!(self.kind, VisibilityKind::Path(_) | VisibilityKind::Crate(_))
+    }
+
+    /// Returns `true` if the visibility is declared as `pub(crate)`. This is a
+    /// special case of the `pub(<path>)` visibility.
+    ///
+    /// This function checks if the visibility is restricted and the defined path
+    /// belongs to the root module of the crate. Meaning, that this can also be `true`,
+    /// if the visibility uses `pub(super)` to travel up to the crate root.
```
Veetaha, line 323: So this will return true if the path refers to the root of the crate. That looks more like the semantic property of this type, not the ast. Maybe it makes sense to make the ast/sem distinction more exlicit? Is it even possible to check if the visibility is `pub(crate)` syntactically (and not `pub(super)` or `pub(in ...)` that lead to the root of the crate).

marker_api/src/ast/item.rs
```diff
@@ -238,42 +240,182 @@ use impl_item_data;
...
+    ///
+    /// ```rs
```
Veetaha, line 325: `rs` disables the code highlighting of in the doc comment in rust-analyzer, and I think `cargo test --doc` will also not try to compile this code as a doc test. Just use bare triple-tick

```suggestion
    /// ```
```

marker_api/src/ast/item.rs
```diff
@@ -238,42 +240,182 @@ use impl_item_data;
...
+    ///
+    ///     // Returns `false` since the item is not visible from the root of the crate.
+    ///     pub(in crate::example_1) fn bar() {}
+    ///
+    ///     // Returns `true` as the visibility is restricted to the root of the crate.
+    ///     pub(crate) baz() {}
+    ///
+    ///     // Returns `true` as the visibility is restricted to `super` which is the
+    ///     // root of the crate.
+    ///     pub(crate) boo() {}
```
Veetaha, line 340:
```suggestion
    ///     pub(super) boo() {}
```

marker_api/src/ast/item.rs
```diff
@@ -238,42 +240,182 @@ use impl_item_data;
...
+    /// }
+    ///
+    /// // Returns `false` since the visibility is not restricted
+    /// fn example_in_root() {}
```
Veetaha, line 344: This makes it again special case. I don't like how syntactic and semantic meaning of `pub(crate)` are intermixed here.
What I would expect is smth like this:
- `ast::Visibility::is_pub_crate()`: checks if the visibility is spelled exactly as `pub(crate)` returning false for any other syntax

- `ast::Visibility::is_pub_in_path()`: checks if the visibility is spelled exactly as `pub(in ...)` returning false for any other syntax

- `sem::Visibility::is_pub_crate()`: checks if the visibility makes the symbol accessible in the crate root (`pub(crate)`, `pub(super)`, `pub(in ...)`, default visibility in `lib.rs` all would return `true`)

- `sem::Visibility::is_pub_in_path(path)`: checks if the visibility makes the symbol accessible under the given path (any syntax e.g. `pub(crate)`, `pub(in ...)` may return `true` only if the resulting resolved visibility allows the symbol to be accessed in `path` module).

marker_api/src/ast/item.rs
```diff
@@ -238,42 +240,182 @@ use impl_item_data;
...
+    /// ```
+    pub fn is_pub_crate(&self) -> bool {
+        matches!(self.kind, VisibilityKind::Crate(_))
+    }
+
+    /// Returns `true` if a visibility has been defined.
+    pub fn is_defined(&self) -> bool {
+        !matches!(self.kind, VisibilityKind::Default(_) | VisibilityKind::DefaultPub)
+    }
```
Veetaha, line 353: The term "defined" implies there is a case where the visibility is "undefined", which reads to me as if "we just don't know what the visibility is if it's not defined". I think `is_non_default` or is `is_explicit` would read better.

marker_api/src/ast/item.rs
```diff
@@ -238,42 +240,182 @@ use impl_item_data;
...
+
+    /// Returns the [`ItemId`] of the module where this item is visible in, if the
+    /// visibility is defined to be public inside a specified path.
+    ///
+    /// See [`Visibility::module_id`] to get the `ItemId`, even if the item or
+    /// field uses the default visibility.
+    pub fn pub_with_path_module_id(&self) -> Option<ItemId> {
```
Veetaha, line 360: This again looks weird due to the mixture of syntactic and semantic properties of the visibility.

I think if we split the representation into `ast` and `sem` there would be only a method on `sem` type:

- `sem::Visibility::visibility_root() -> Option<ItemId>`: returns the `ItemId` of the root module this visibility level allows access in; returns `None` if the visibility is `pub`

And there wouldn't be such a method for `ast::Visibility`. I think it would be more straightforward to just expose the syntactic `VisibilityKind` to them, and the users would then handle all possible syntactic flavors of the visibility they want to check for.
But the `VisibilityKind` in the `ast` module would look like this:

```rust
enum VisibilityKind {
  Pub,
  PubIn(Path),
  PubCrate,
  
  /// There is no distinction between `Default/DefautPub` because in the synatctic representation
  /// we only care if the visibility was explicitly specified or not
  Default, 
}
```

Maybe I'm dreaming here, and rustc doesn't provide the original syntax form of the visibility, in which case it would be a bummer.

**UPD**: I noticed the `module_id` method only after writing this comment.

marker_rustc_driver/src/conversion/marker/ast/item.rs
```diff
@@ -159,6 +165,26 @@ impl<'ast, 'tcx> MarkerConverterInner<'ast, 'tcx> {
         Some(item)
     }
 
+    fn to_visibility(&self, owner_id: hir::def_id::LocalDefId, vis_span: rustc_span::Span) -> Visibility<'ast> {
+        let span = if vis_span.is_empty() {
+            None
+        } else {
+            Some(self.to_span_id(vis_span))
+        };
```
Veetaha, line 173:
```suggestion
        let span = (!vis_span.is_empty()).then(|| self.to_span_id(vis_span));
```

marker_rustc_driver/src/conversion/marker/ast/item.rs
```diff
@@ -249,7 +275,7 @@ impl<'ast, 'tcx> MarkerConverterInner<'ast, 'tcx> {
             // field after the next sync. See #55
```
Veetaha, line 275: This fixme comment may be removed already, right?

marker_uilints/src/lib.rs
```diff
@@ -144,6 +149,26 @@ impl LintPass for TestLintPass {
                         diag.note(format!("Body: {:#?}", cx.ast().body(func.body_id().unwrap())));
                     });
             }
+            if matches!(
+                item.ident().map(marker_api::span::Ident::name),
+                Some(name) if name.starts_with("test_vis")
+            ) {
```
Veetaha, line 155:
```suggestion
            if item.ident().is_some_and(|ident| ident.name().starts_with("test_vis")) {
```

marker_api/src/ast/item.rs
```diff
@@ -238,42 +240,182 @@ use impl_item_data;
...
+    pub fn new(id: ItemId, span: SpanId, vis: Visibility<'ast>, ident: Ident<'ast>) -> Self {
+        Self { id, span, vis, ident }
     }
 }
 
-/// This struct represents the visibility of an item.
+/// The declared visibility of an item or field.
+///
+/// Note that this is only the syntactic visibility. The item or field might be
+/// reexported with a higher visibility, or have a high default visibility.
```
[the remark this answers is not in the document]
xFrednet: That's really interesting, I never knew. Thank you!

xFrednet:
This is still WIP, so far I've only applied the easy changes and skimmed over your suggestions. I'll answer properly soon. Thank you for the review!

marker_api/src/ast/item.rs
xFrednet: Rustc sadly erases the information how the path has been specified. That's why this type provides semantic information when it comes to this. I could rearrange it, so that we have a semantic visibility, and the AST visibility, basically just wraps that one and optionally provides a `Span`. What do you think about this structure?

marker_api/src/ast/item.rs
xFrednet: See: https://github.com/rust-marker/marker/pull/294#discussion_r1375321200

marker_api/src/ast/item.rs
Veetaha: I see, that's a pity. Anyway, since we don't have a different from the original AST representation of the visibility, I think it does make sense

xFrednet:
Okay, I've rebased. Only the last two commits are new. I've now split the visibility into a semantic and ast representation, as discussed previously. Sorry that it took so long, to get this fixed :sweat_smile: Hope you have a beautiful weekend :D

marker_api/src/sem/item.rs
```diff
@@ -0,0 +1,160 @@
...
+        matches!(self.kind, VisibilityKind::Path(_))
+    }
+
+    /// Returns `true` if the visibility is declared as `pub(crate)`. This is a
+    /// special case of the `pub(<path>)` visibility.
+    ///
+    /// This function checks if the visibility is restricted and the defined path
+    /// belongs to the root module of the crate. Meaning, that this can also be `true`,
+    /// if the visibility uses `pub(super)` to travel up to the crate root.
+    // Ignore, since the `in crate::example_1` path doesn't work for doc tests
```
Veetaha, line 64: I wonder how this code is layed out during compilation. That's super-unintuitive.
I've checked some metadata:
```
[marker_api/src/sem/item.rs:27] file!() = "marker_api/src/sem/item.rs"
[marker_api/src/sem/item.rs:28] module_path!() = "rust_out"
[marker_api/src/sem/item.rs:8] module_path!() = "rust_out::example_1"
thread 'main' panicked at marker_api/src/sem/item.rs:30:1:
rust_out::main::_doctest_main_marker_api_src_sem_item_rs_64_0::example_1::Baz
```
The panic mentions the `type_name` of a struct defined inside of an `example_1`. I guess rustdoc generates just a single crate for many dotctests if the code example doesn't include an explicit `main` function.

However, if you add an explicit `main()` function the path to `Baz` becomes this:
```
thread 'main' panicked at marker_api/src/sem/item.rs:4:5:
rust_out::example_1::Baz
```

So to remove the `ignore` in this doctest you can add an explicit `# fn main() {}` somewhere here. This will make `crate::example_1` work.

marker_api/src/sem/item.rs
```diff
@@ -0,0 +1,160 @@
...
+    ///
+    ///     // Returns `true` as the visibility is restricted to the root of the crate.
+    ///     pub(crate) fn baz() {}
+    ///
+    ///     // Returns `true` as the visibility is restricted to `super` which is the
+    ///     // root of the crate.
+    ///     pub(super) fn boo() {}
+    /// }
+    ///
+    /// // Returns `false` since the visibility is not restricted
```
Veetaha, line 83: Technically, this item is also `pub(crate)`. That's a bit unintuitive. Could we identify that the default visibility at the root of the crate as implicitly `pub(crate)` as well? I guess it would require just another `VisibilityKind`

marker_api/src/sem/item.rs
```diff
@@ -0,0 +1,160 @@
...
+    /// fn example_in_root() {}
+    /// ```
+    pub fn is_pub_crate(&self) -> bool {
```
Veetaha, line 86: I think people will be confused if we name this method after the special `pub(crate)` syntax when this method covers more than just that.

```suggestion
    pub fn is_crate_scoped(&self) -> bool {
```

Given the amount of info consider this number of methods instead:
- `is_crate_scoped() -> bool` (true if `scope` resolves to the current crate no matter what syntax was used)
- `is_pub() -> bool { self.scope().is_none() }`
- `is_default() -> bool`
- `scope() -> Option<ItemId>`
If the users wants to check if the visibility was specified (non-default), they would use `!is_default()`. If they want to exclude `pub` they will use `!is_default && !is_pub()`.

In other words I'd like to make it explicit that the semantic visibility no longer represents the original syntax that was used.

You can also notice that in the list of methods I suggested there isn't `pub_with_path_module_id`. I think it's not needed because it's expressible with just `vis.scope().filter(|_| !vis.is_default())` or similar.

Veetaha (commented):
Consider making design a bit more minimalistic but obvious as suggested in my comment.

marker_api/src/sem/item.rs
xFrednet: That works perfectly, thank you! :heart:

marker_api/src/sem/item.rs
xFrednet: Yes we could, and I would actually guess that this is currently wrongly classified as `pub(crate)`.

marker_api/src/sem/item.rs
xFrednet: I really like the suggested names. I'll make the changes. We should be able to reparse at least part of the text snippet. I haven't worked with Rustc's lexer and parser yet, which is partially why I'm avoiding such solutions :sweat_smile:

xFrednet:
Okay, this should hopefully be it. Thank you for the excellent review. I think this very is way better, than my initial proposal :)

marker_api/src/sem/item.rs
```diff
@@ -80,57 +75,52 @@ impl<'ast> Visibility<'ast> {
...
-    /// increase the visibility.
-    pub fn module_id(&self) -> Option<ItemId> {
+    pub fn scope(&self) -> Option<ItemId> {
         match self.kind {
-            VisibilityKind::Path(id) | VisibilityKind::Crate(id) | VisibilityKind::Default(id) => Some(id),
+            VisibilityKind::Path(id)
+            | VisibilityKind::Crate(id)
+            | VisibilityKind::DefaultCrate(id)
+            | VisibilityKind::Default(id) => Some(id),
             _ => None,
```
Veetaha, line 124: `[nit]`
I'd prefer an exhaustive match to make sure all cases are handled if `VisibilityKind` enum is modified

marker_api/src/sem/item.rs
Veetaha: Re-parsing the text snippet is more of a hack. I wonder how clippy fights with this? Maybe other kinds of lint passes such as `EarlyLintPass` have the necessary ast info?

Veetaha (approved):
LGTM. Design is hard, but things should go on 👍

marker_api/src/sem/item.rs
xFrednet: Clippy usually just checks if things are public, and that's it. In Clippy we don't really have lints that span over more than a module and those that do don't require the visibility AFAIK.

For the `EarlyLintPass`, I've just checked and that one still has the path [[1](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/enum.VisibilityKind.html)]. The big problem with this approach is that nodes from the AST, are hard to connect to the nodes from the HIR, as the AST doesn't have a unique ID for everything. `HirId`s are also structured differently.

The current rustc interface has covered all of Clippy's use cases so far. So I hope that this is enough for Marker.

PR #299, opened 2023-10-22
Fix archive layout for marker_rustc_driver on ARM

Veetaha:
The layout of `marker_rustc_driver-aarch64-unknown-linux-gnu.tar.gz` and `zip` turns out to differ from the layout of all other artifacts. This is because this version of driver is compiled separately using a custom quemu setup, packaging and upload and so it went slightly wrong.

The difference is that the file inside of the archive contains the architecture suffix, but it shouldn't.

I've manually updated the current 0.3.0 release by downloading the marker_rustc_driver ARM build, decompressing it, renaming the file to `marker_rustc_driver`, and running the part of `upload.sh` script that does the packaging.

### DIFF
--- a/.github/workflows/release-on-tag.yml
+++ b/.github/workflows/release-on-tag.yml
@@ -156,9 +156,9 @@ jobs:
           run: |
             export PATH="/root/.cargo/bin:$PATH"
             cargo build -p marker_rustc_driver --release
-            cp target/release/marker_rustc_driver /artifacts/${{ env.artifact }}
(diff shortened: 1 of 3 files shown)


xFrednet (approved):

xFrednet:
LGTM, thank you for the updated. From this change and the issues you created, is it right to assume that you tried implementing a few lints in Marker?

Veetaha:
Yes, I implemented several 5 lints in a private repository at work

PR #306, opened 2023-10-31
API: Use a `ModItem` as the root of a crate

### LINKED ISSUE #279: [Feat]: Add method to check the root module of  a crate
### Summary

Marker currently models a crate as a `crate` object with holds several items. `marker_adapter` then visits every item and passes it to the lint crates. This setup currently doesn't pass the crate object to the lint crates. I imagine the following:

1. Add a `check_crate(cx: &MarkerContext<'ast>, krate: ast::Crate<'ast>)` method to the `LintPass` crate
2. Remodel the crate object, to hold a single module which holds all the items in the root of the crate.
    * This is more inline with rustc and also what I would have expected.

These changes will "break" the API of the `Crate` struct. However, in this case that's totally fine, since there is currently no way to actually get a `Crate` instance inside lint crates. (Hence this issue xD)

xFrednet:
This PR has 3 major changes:

1. The items of a `Crate` are now wrapped in a `ModItem`, that is the root module of the crate.
2. The `LintPass` trait now as a new `check_crate` method.
3. Rustc's driver no longer ICEs on spans from compiler generated code. 

---

It has been some time, but I'm back. Exams really took it out of me, and then I felt so exhausted, that I needed to take a step back for some time, just to breathe. This should hopefully be the kickoff for me getting back into things :D

---

Closes #279 

Feedback is as always highly appreciated! And for everyone reading this: I hope you have a beautiful day. It's awesome that you're a part of open source, be proud of yourself! Life can be stressful, so don't forget to take a break every once in a while.


marker_rustc_driver/src/conversion/marker.rs
```diff
@@ -304,14 +307,37 @@ impl<'ast, 'tcx> MarkerConverterInner<'ast, 'tcx> {
...
+
+        let krate = self.alloc(
+            Crate::builder()
+                .id(self.to_crate_id(hir::def_id::LOCAL_CRATE))
+                .root_mod(self.local_crate_mod())
+                .build(),
+        );
+        self.krate
+            .set(krate)
+            .expect("this can't be filled, since we checked earlier and this stage is single threaded");
```
Veetaha, line 323: There is [get_or_init](https://doc.rust-lang.org/std/cell/struct.OnceCell.html#method.get_or_init) that allows to avoid unwraps

marker_rustc_driver/src/conversion/marker/span.rs
```diff
@@ -25,7 +25,13 @@ impl<'ast, 'tcx> MarkerConverterInner<'ast, 'tcx> {
         let ctxt = rust_span.ctxt();
 
         if !ctxt.is_root() {
-            return SpanSource::Macro(self.alloc(self.to_expn_info(&ctxt.outer_expn_data())));
+            let expn_data = ctxt.outer_expn_data();
+            return match expn_data.kind {
+                rustc_span::ExpnKind::Macro(_, _) => SpanSource::Macro(self.alloc(self.to_expn_info(&expn_data))),
+                rustc_span::ExpnKind::AstPass(_) => SpanSource::Buildin(self.buildin_span_source),
+                rustc_span::ExpnKind::Desugaring(_) => unreachable!("desugaring spans should never be crated"),
```
Veetaha, line 32:
```suggestion
                rustc_span::ExpnKind::Desugaring(_) => unreachable!("desugaring spans should never be created"),
```

marker_uilints/tests/ui/item/print_me_root_module.stderr
```diff
@@ -0,0 +1,108 @@
...
+                  },
+                  items: [
+                      Use(
+                          UseItem {
+                              data: CommonItemData {
+                                  id: ItemId(..),
+                                  span: SpanId(..),
+                                  vis: Visibility {{ /* WIP: See rust-marker/marker#26 */}},
+                                  ident: Ident {
+                                      name: "",
```
Veetaha, line 22: This ident is suspiciously empty.shouldn't it be None?

marker_uilints/tests/ui/item/print_me_root_module.stderr
```diff
@@ -0,0 +1,108 @@
...
+                                      span: $DIR/print_me_root_module.rs:1:1 - 1:1,
+                                  },
+                              },
+                              use_path: AstPath {
```
Veetaha, line 26: I'd be glad if lints didn't even need to see the standard prelude. I guess it deserves it's own ugly crutch to skip such code. And.. in general, I think lints don't ever need to see code with the "built-in" spans auto-generated by the compiler. I imagine someone writes a lint that triggers on a standard prelude use item and thus is cursed to never be able to ignore that false positive unless ignoring it at crate-level meaning the lint is just disabled and *use*less 😺

marker_api/src/interface.rs
```diff
@@ -78,6 +79,12 @@ macro_rules! export_lint_pass {
                 extern "C" fn info() -> $crate::LintPassInfo {
                     super::__MARKER_STATE.with(|state| state.borrow_mut().info())
                 }
+                extern "C" fn check_crate<'ast>(
```
Veetaha, line 82: I wonder if this PR also changes the behaviour of "check_item"? I think that method must also traverse the root crate module. In general, I think `check_crate` will be very rarely needed, except for lints that specially handle the crate's root module. Anyhow, the crate's root module should not be anything special in the context of `check_item`

marker_api/src/span.rs
```diff
@@ -235,6 +235,7 @@ impl<'ast> std::fmt::Debug for Span<'ast> {
                 fmt_pos(file.try_to_file_pos(self.end))
             ),
             SpanSource::Macro(expn) => format!("[Inside Macro] {:#?}", expn.call_site()),
+            SpanSource::Buildin(_) => "[From Prelude]".to_string(),
```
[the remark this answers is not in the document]
xFrednet: Ahh right, that's why the spelling correction complained :sweat_smile:

marker_uilints/tests/ui/item/print_me_root_module.stderr
xFrednet: Skipping the import items from the prelude would be pretty simple, items already have an ugly crunch to handle such things.

The question for me, is more if we want this? So these items are definitely not intended for lint emissions, but they are required to get the full information about what namespace etc are available. On the other hand, these changes are always bound to Editions, so if we provide information about that, it should be sufficient to get all the information needed for linting :thinking:

marker_api/src/interface.rs
xFrednet: The root module is passed to `check_item` like any other module. The goal was to handle the root module, like a normal module, to make linting easy.

marker_uilints/tests/ui/item/print_me_root_module.stderr
Veetaha: The namespace information should be queried from the `TyCtxt`. I don't think that lint authors will specifically traverse all uses and do manually name resolution, which is a complex and recursive algorithm.

marker_uilints/tests/ui/item/print_me_root_module.stderr
xFrednet: I'm not quite sure how this value was constructed. I've now removed all compiler generated items from Marker's AST representation and this "fixed" (meaning hid this issue) for now. I believe I have seen some circumstances, where the identifier was empty of something in the compiler for a valid reason, but I don't remember where. As for the `None` part. The `Ident` struct is used a lot, adding an option in it, would make handling it very cumber sum. So, even if this was a bug, I think it's better to have `""` as a symbol than using an option inside the `Ident`

marker_uilints/tests/ui/item/print_me_root_module.stderr
xFrednet: Done :+1:

marker_api/src/span.rs
```diff
@@ -235,7 +235,7 @@ impl<'ast> std::fmt::Debug for Span<'ast> {
                 fmt_pos(file.try_to_file_pos(self.end))
             ),
             SpanSource::Macro(expn) => format!("[Inside Macro] {:#?}", expn.call_site()),
-            SpanSource::Buildin(_) => "[From Prelude]".to_string(),
+            SpanSource::Builtin(_) => "[From Prelude]".to_string(),
```
Veetaha, line 238: If lints can't observe the prelude, then we may go without the "Builtin" span source? Or is there any other way this "Builtin" span source can be created? The test that was added in this PR no longer contains such a span source. If there is any other way it could be created, it would be good to have a test that showcases how it can be constructed

xFrednet: The filter mechanism added in this PR should filter out all spans with a builtin source. However, I'm not 100% sure if that really removes every trace. Before this new `SpanSource` Marker would ICE when requesting these spans. I'm in favor of keeping it in the API as it will prevent potential ICEs. If we're starting to look at the stabilization of Marker, we can review if this value ever came up again.

I can imagine that also some unstable features might fall back on this source, during initial development. This is just a guess, though.

I'll change the text of this function to say `"[Builtin]"` instead.

Veetaha (approved):

Issue #274, opened 2023-10-05
Release post-mortem


Veetaha:
0.3.0 release was done, but there were a number of problems with it.
- The `crates-io` publishing job failed. I wonder why is that. It managed to publish part of the crates, but choked on `marker_utils` complaining that the publishing identity doesn't have rights to publish that package. @xFrednet could you elaborate how that could happen and what measures were taken to remediate this?
- The `release` CI job couldn't push commits to `master` due to the merge queue enabled on the repo. I didn't test this process with a merge queue enabled in the repo, and I see that it was fixed. `git push --force-with-lease` didn't resolve the issue and should be removed from the CI template. I suppose you tweaked the repo settings somehow to unblock this?

xFrednet:
> The crates-io publishing job failed. 

I forgot to add the `rust-marker/crate-owners` group to the `marker_utils` crate. :sweat_smile:  That's fixed now. I decided to push the remaining packages myself for this release, as I was uncertain how `cargo release` would handle pushing crate versions that have already been published.

> I suppose you tweaked the repo settings somehow to unblock this?

I tried several configurations, but it seems like the "Require pull request", "Require status check" and "Require merge queues" block pushes completely, even force pushes. One reason could be, that `rust-marker-ci` only has write access and is not registered as a repo admin. On the other hand, I've listed `rust-marker-ci` as a user that's allowed to force push.

Anyways, my non-elegant solution was to just remove those branch protection options completely, run the action and then add them back. Not sure how to solve this properly. Maybe we have to go through PRs and wait for them to be manually merged as well? :thinking: 

---

There are two more things I've noticed:

1. The "[🚧 Development v0.4.0-dev](https://github.com/rust-marker/marker/commit/4c755f9304fbf837c67e0f715fc1aadab9a16703)" commit had a red CI. This actually makes sense, since the "GitHub Action" test tries to fetch the binaries from release, but they haven't been compiled yet.

   I simply rerun the action, once all binaries were uploaded. This made master green again. I don't see a simple way to solve this. We can't really wait with the pushes, until the binaries are online, since the binaries are build with those commits. Rerunning the CI was easy, just something to keep in mind :)
   
2. Not related to the github release process, but more releases in general. I noticed a bug in `marker_uitest` shortly before the release, while playing around with the `marker-example-lints` repo. I think it would be good to define some smoke tests (And maybe partially automate them) to run before every release. It could be a simple:
[...]

PR #308, opened 2023-11-08
Improve retries in the installation scripts

Veetaha:
Today on our CI the installation of marker randomly failed:
image

GitHub returned 403 due to some internal bugs I suppose. To fix this I added `--retry-all-errors` which should retry in such cases and added more configurations for retries. This way on CI we use the retry config that gives up in 30 minutes. This solves the problem of red master CI when doing a release (part of #274).


xFrednet (approved):

xFrednet:
Everything looks good to me, thank you for the update :D

Issue #297, opened 2023-10-19
Use consistent naming `fn pat()` vs `fn pattern()`


Veetaha:
Some methods in `marker_api` are called [`pat()`](https://docs.rs/marker_api/latest/marker_api/?search=pat), but there are few that are called [`pattern`](https://docs.rs/marker_api/latest/marker_api/?search=pattern). We need to make the naming consistent. Maybe move these methods under a trait and/or add a custom lint to `marker_lints` that checks that we don't call the term "pattern" inconsistently.

Issue #295, opened 2023-10-19
Rename `CallExpr::operand` to `CallExpr::func`


Veetaha:
Operand is the "argument" for the operator, thus having both `operand()` and `args()` looks confusing. The `operator` is what the thing that is invoked, it seems to me like an oversight in terminology.

https://docs.rs/marker_api/latest/marker_api/ast/struct.CallExpr.html#method.operand

xFrednet:
The official docs call the function the `operand`

> A call expression calls a function. The syntax of a call expression is an expression, called the *function operand*, followed by a parenthesized comma-separated list of expression, called the *argument operands*. 
>
> ~https://doc.rust-lang.org/reference/expressions/call-expr.html#call-expressions

Which is why I've chosen it as the official name, but I see the possible confusion. I'm fine with changing it, if this makes the API clearer

Veetaha:
The official docs call the function the `operand`, because they treat `()` as the `operator`, in which case it has two "operands": the function and its arguments list. So both of them are "operands".

I propose to follow the naming that `syn` uses for [call expressions](https://docs.rs/syn/latest/syn/struct.ExprCall.html) meaning `func` and `args`.

PR #309, opened 2023-11-14
API: Rename `CallExpr::operand` -> `CallExpr::func`;  `*::pattern` -> `*::pat`

xFrednet:
Some smaller breaking changes before `v0.4.0` probably best summarized by the changelog entry:

- Renamed `CallExpr::operand` to `CallExpr::func`
- Renamed `RefPat::pattern()` -> `RefPat::pat()`
- Renamed `OrPat::patterns()` -> `RefPat::pats()`

Yeah, and that's it :D

---

I also moved the semver check higher in our CI file. I thought about making it the first item, but I feel like the others are still more important and should be the first visible CI checks. Having semver higher up is already a nice change :)

---

Closes https://github.com/rust-marker/marker/issues/297

Closes https://github.com/rust-marker/marker/issues/295

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,33 @@ jobs:
       - uses: actions-rust-lang/setup-rust-toolchain@v1
       - run: cargo fmt --check
 
+  # This check is allowed to fail, while Marker is still unstable. It's mostly
+  # a sanity check, to make sure we don't have any unintentional API breakage :)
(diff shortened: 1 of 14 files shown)


Veetaha (approved):
I recommend adding "Closes #297" and "Closes #295" to the PR description to make the issues auto-closed when PR is merged

xFrednet:
Good call, I forgot about that. Thanks!

PR #317, opened 2023-11-22
Fix marker_rustc_driver build ARM just one more time

Veetaha:
When last time I fixed the build of marker_rustc_driver on ARM I tested my changes only locally in an isolated folder where I manually "unit-tested" the `upload.sh` script. Of course, when the build happens on CI the folders layout is entirely different. That's why the ARM packaging script broke but in other way.

On CI we mount the `.` fodler as `/artifacts` in `quemu`, but `.` folder includes the `marker_rustc_driver` crate source code folder. So when we do `cp target/release/marker_rustc_driver /artifacts/marker_rustc_driver` we copy the binary inside of that folder...

I fixed this by mounting `./artifacts` which is guaranteed to be an empty folder.

Also manually fixed and reuploaded the artifacts in 0.4:
image

This time I tested it end-to-end in my fork repo

### DIFF
--- a/.github/workflows/release-on-tag.yml
+++ b/.github/workflows/release-on-tag.yml
@@ -146,7 +146,7 @@ jobs:
         with:
           arch: aarch64
           distro: ubuntu20.04
-          dockerRunArgs: --volume .:/artifacts
+          dockerRunArgs: --volume ./artifacts:/artifacts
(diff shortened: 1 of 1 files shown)


xFrednet (approved):

xFrednet:
Looks good to me, thank you!

Issue #318, opened 2023-11-23
[Bug]: `cargo marker setup --auto-install-toolchain` fails with E0554 `#[feature]` on stable channel


xFrednet:
### Summary

I've extracted this error from https://github.com/trailofbits/dylint/issues/839, there it was reported by @tigerros

> @xFrednet Yes, I want to make something like Clippy, just for a specific framework. I think Marker will do nicely. I'm having some issues with installation though. I get this output when running `cargo marker setup --auto-install-toolchain`. I have `rustc`, `rustc-dev` and `llvm-tools` installed.

### Version

```text
Marker v0.4
```

### Logs

```text
> cargo marker setup --auto-install-toolchain
> info: syncing channel updates for 'nightly-2023-10-05-x86_64-pc-windows-msvc'
> 724.5 KiB / 724.5 KiB (100 %) 573.5 KiB/s in  1s ETA:  0s
> info: latest update on 2023-10-05, rust version 1.75.0-nightly (2bbb61989 2023-10-04)
> info: component 'llvm-tools' for target 'x86_64-pc-windows-msvc' is up to date
> info: component 'rustc-dev' for target 'x86_64-pc-windows-msvc' is up to date
> 
>   nightly-2023-10-05-x86_64-pc-windows-msvc unchanged - rustc 1.75.0-nightly (2bbb61989 2023-10-04)
> 
> info: checking for self-update
> 
>       Marker compiling rustc driver v0.4.0 with nightly-2023-10-05
>     Updating crates.io index
>   Installing marker_rustc_driver v0.4.0
>     Updating crates.io index
>    Compiling proc-macro2 v1.0.69
>    Compiling unicode-ident v1.0.12
>    Compiling windows_x86_64_msvc v0.48.5
>    Compiling cfg-if v1.0.0
>    Compiling once_cell v1.18.0
>    Compiling cc v1.0.83
>    Compiling winapi v0.3.9
>    Compiling thiserror v1.0.50
>    Compiling rustc-demangle v0.1.23
>    Compiling lazy_static v1.4.0
>    Compiling tracing-core v0.1.32
>    Compiling sharded-slab v0.1.7
>    Compiling windows-targets v0.48.5
>    Compiling thread_local v1.1.7
>    Compiling windows-sys v0.48.0
>    Compiling quote v1.0.33
>    Compiling serde v1.0.193
>    Compiling pin-project-lite v0.2.13
>    Compiling syn v2.0.39
[...]

PR #319, opened 2023-11-24
Cargo: Fix compiling driver from sources outside of the marker repo on Windows

Veetaha:
Closes https://github.com/rust-marker/marker/issues/318

I immediately identified the problem by the [logs](https://github.com/rust-marker/marker/issues/318#issuecomment-1825917130) because this bug is specific to Windows only.

It was described as "Problem 2" in the description of the PR https://github.com/rust-marker/marker/pull/217, although that appeared in a bit different place at that moment.

This is just a quick fix, it doesn't setup testing on CI. I just reproduced the problem manually on my windows laptop and fixed it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,14 @@ The following components are considered to be internal and they are excluded fro
 
 ## [Unreleased]
 
-The [v0.5.0 milestone] contains a list of planned changes.
+## [0.4.1] - 2023-11-24
 
-[v0.5.0 milestone]: https://github.com/rust-marker/marker/milestone/5
+[#319]: https://github.com/rust-marker/marker/pull/319
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


xFrednet (approved):
The change looks good to me. Only rustfmt is complaining. Let's do a patch release after this, since there hasn't been anything on master since the last release. (Unless you thinkg it's not worth it)

Veetaha:
I think it makes sense to do a patch release to al least test this workflow.
I'll update the changelog

xFrednet:
I'll trigger the release workflow, since it still seems to require some changes to the branch protection rules

Veetaha:
I don't think it requires, try without changes

xFrednet:
Too late, I'll do it the next time. This should also fix the ARM archives :+1:

PR #321, opened 2023-11-25
Remove params that aren't supported by `cargo release publish`

Veetaha:
I was confused by [this warning on CI](https://github.com/rust-marker/marker/actions/runs/6984551469/job/19007537021#step:6:14)

It looked to me as if cargo release wanted to do a commit and push changes to the remote. I still don't understand why that warning is emitted because `cargo release` is used only for publishing. Maybe that's a micro-bug in `cargo-release`. Anyway, this time it was me who broke crates publishing. Hopefully, this was the last obstacle

### DIFF
--- a/.github/workflows/release-on-tag.yml
+++ b/.github/workflows/release-on-tag.yml
@@ -199,6 +199,6 @@ jobs:
       # No idea why this happens, but let's just keep a single cargo release
       # to avoid this issue and to simplify the workflow, because the problem
       # we tried to protect against is not that likely to happen anyway.
-      - run: cargo release publish --no-confirm --no-verify --no-push --no-tag --allow-branch '*'
+      - run: cargo release publish --no-confirm --no-verify --allow-branch '*'
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
--- a/scripts/download/cargo-release.sh
+++ b/scripts/download/cargo-release.sh
@@ -6,7 +6,7 @@ script_dir=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
 
 . $script_dir/lib.sh
 
-version=0.24.11
+version=0.25.0
 
 base_url=https://github.com/crate-ci/cargo-release/releases/download/v$version
 


Veetaha:
Btw, I think you should release the 0.4.2 crates manually, just one more time to make the `github-action-test` jobs pass. Now they fail on:
```
  × Lint crate marker_lints uses the version of marker_api 0.4.1 that is incompatible with the version of marker_api 0.4.2 used in the
  │ driver
  help: update either the marker_api dependency in marker_lints or update the driver to the latest version
```

xFrednet:
Hmm, I've installed `cargo-release v0.25.0` by running `cargo install cargo-release`, checkout out the v0.4.2 tag and ran the command from this PR and got an error:

```
$ cargo release publish --no-confirm --no-verify --allow-branch '*'
warning: push target `origin/HEAD` doesn't exist
  Publishing marker_error
    Updating crates.io index
   Packaging marker_error v0.4.2 (/home/xfrednet/workspace/rust/marker/marker_error)
    Packaged 7 files, 58.3KiB (12.7KiB compressed)
   Uploading marker_error v0.4.2 (/home/xfrednet/workspace/rust/marker/marker_error)
warning: aborting upload due to dry run
  Publishing cargo_marker
    Updating crates.io index
   Packaging cargo_marker v0.4.2 (/home/xfrednet/workspace/rust/marker/cargo-marker)
    Updating crates.io index
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for the requirement `marker_error = "^0.4.2"`
  candidate versions found which didn't match: 0.4.1, 0.4.0, 0.3.0, ...
  location searched: crates.io index
  required by package `cargo_marker v0.4.2 (/home/xfrednet/workspace/rust/marker/cargo-marker)`
  if you are looking for the prerelease package it needs to be specified explicitly
      marker_error = { version = "0.3.0-dev" }
  perhaps a crate was updated and forgotten to be re-vendored?
```

The reference to `marker_error = { version = "0.3.0-dev" }` confuses me. I also can't find any references to it, when searching for `0.3.0` or `marker_error` in the repo. Do you have an idea, where this could come from?

Veetaha:
The line `marker_error = { version = "0.3.0-dev" }` is just a useless "help message". The actual error is:
```
  failed to select a version for the requirement `marker_error = "^0.4.2"`
  candidate versions found which didn't match: 0.4.1, 0.4.0, 0.3.0, ...
```
Maybe the "dry run" mode of cargo release doesn't take into account that this package would be uploaded as part of the same release command. Probably because `cargo package` requires the crates to be published during packaging

xFrednet:
Yup, everything worked now. Guess this leaves the question, if we want to update the `cargo-release` version of our CI to v0.25^. We can also merge the PR as is :)

Veetaha:
Updated `cargo release` version, thanks

.github/workflows/release-on-tag.yml
```diff
@@ -199,6 +199,6 @@ jobs:
       # No idea why this happens, but let's just keep a single cargo release
       # to avoid this issue and to simplify the workflow, because the problem
       # we tried to protect against is not that likely to happen anyway.
-      - run: cargo release publish --no-confirm --no-verify --no-push --no-tag --allow-branch '*'
+      - run: cargo release publish --no-confirm --no-verify --allow-branch '*'
```
xFrednet, line 202: I think this version requires the `--execute` flag, to actually execute the publish command. At least it required that, when I ran it locally

Veetaha: Right, I'll fix it in a separate PR

Veetaha: Created https://github.com/rust-marker/marker/pull/323

Issue #141, opened 2023-06-16
API: Add a method to check if a specific type implements a trait


xFrednet:
This method could look similar to Clippy's [`implements_trait`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/ty/fn.implements_trait.html) util.

Once #93 is resolved, we should have access to the `TyDefId` of traits. Semantic paths are also almost done (#139). This function will require an option to have the user specify generic arguments for the trait.

In short: This will be *fun* or at least complex, but the base for this should be set with the two linked issued.

xFrednet:
Not sure, if I'll be able to complete this for v0.2.0, but I've added it to the tracker for now :)

PR #322, opened 2023-11-25
API: Add impl `From<*>` for `sem::TyKind` and correct some names of semantic types

xFrednet:
This PR adds boilerplate code required for the implementation of https://github.com/rust-marker/marker/issues/141. I've also changed all semantic type construction, to use the `TypedBuilder` instead of the old `new()` function.

This PR also includes some changes, visible to API consumers, probably best summarized by the changelog entires:

- `sem::TyKind` now implements `From<*>` for all semantic types.
- Renamed `sem::TyKind::FnTy` -> `sem::TyKind::Fn`
- Renamed `sem::TyKind::ClosureTy` -> `sem::TyKind::Closure`

That's basically it. Some boring repetitive code changes, which will hopefully make it easier to expand semantic types in the future :D

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ The following components are considered to be internal and they are excluded fro
 
 - [#322]: Renamed `sem::TyKind::FnTy` -> `sem::TyKind::Fn`
 - [#322]: Renamed `sem::TyKind::ClosureTy` -> `sem::TyKind::Closure`
+- [#322]: Renamed `sem::ClosureTy::closure_ty_id` -> `sem::ClosureTy::def_id`
 
 ## [0.4.2] - 2023-11-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-[Unreleased]: https://github.com/rust-marker/marker/compare/v0.4.0...HEAD
+[Unreleased]: https://github.com/rust-marker/marker/compare/v0.4.2...HEAD
 [0.4.2]: https://github.com/rust-marker/marker/releases/tag/v0.4.2
 [0.4.1]: https://github.com/rust-marker/marker/releases/tag/v0.4.1
 [0.4.0]: https://github.com/rust-marker/marker/releases/tag/v0.4.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 [Unreleased]: https://github.com/rust-marker/marker/compare/v0.4.0...HEAD
+[0.4.2]: https://github.com/rust-marker/marker/releases/tag/v0.4.2
+[0.4.1]: https://github.com/rust-marker/marker/releases/tag/v0.4.1
 [0.4.0]: https://github.com/rust-marker/marker/releases/tag/v0.4.0
 [0.3.0]: https://github.com/rust-marker/marker/releases/tag/v0.3.0
 [0.2.1]: https://github.com/rust-marker/marker/releases/tag/v0.2.1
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 [Unreleased]: https://github.com/rust-marker/marker/compare/v0.4.0...HEAD
+[0.4.2]: https://github.com/rust-marker/marker/releases/tag/v0.4.2
+[0.4.1]: https://github.com/rust-marker/marker/releases/tag/v0.4.1
```
Veetaha, line 3: My bad, I wish someday this becomes autogenerated

Btw. the first line with `Unreleased` link should be updated to `0.4.2` as well

xFrednet:
The semver-check in our CI only found the two expected enum variant renames, I guess this should be good to go :D

Some to think about #141

marker_api/src/sem/ty/fn_ty.rs
```diff
@@ -3,11 +3,17 @@ use crate::{
     sem::generic::GenericArgs,
 };
 
+use super::CommonTyData;
+
+// TODO: breaking change remove Ty from Enum
```
Veetaha, line 8: Is this comment still needed?

marker_api/src/sem/ty/fn_ty.rs
```diff
@@ -38,18 +39,20 @@ impl<'ast> FnTy<'ast> {
...
+    ty_id: TyDefId,
     generics: GenericArgs<'ast>,
 }
 
 impl<'ast> ClosureTy<'ast> {
-    /// This returns the [`ItemId`] of the identified function.
+    /// This returns the [`ItemId`] of the struct that was generated for this closure.
     pub fn closure_ty_id(&self) -> TyDefId {
-        self.closure_ty_id
+        self.ty_id
```
Veetaha, line 52: Consider renaming the method as well?

marker_api/src/sem/ty.rs
```diff
@@ -13,6 +15,8 @@ pub use sequence_ty::*;
 pub use trait_ty::*;
 pub use user_ty::*;
 
+use crate::common::DriverTyId;
```
Veetaha, line 18: Somehow the uses landed in different spots: one before `pub use`s and one after. Let's put them all after `pub use`s because I think this is where Rust analyzer would put them by default.

marker_api/src/sem/ty.rs
```diff
@@ -110,3 +114,43 @@ impl<'ast> TyKind<'ast> {
...
+    pub fn driver_id(&self) -> DriverTyId {
+        self.driver_id
+    }
+}
+
+macro_rules! impl_ty_data {
+    ($self_ty:ty, $enum_name:ident) => {
+        #[cfg(feature = "driver_api")]
+        impl<'ast> $self_ty {
+            pub fn data(&self) -> &$crate::sem::ty::CommonTyData<'ast> {
```
Veetaha, line 144: I'd expect this to fail to compile because this is a `pub` function which returns a `pub(crate)` type.

marker_api/src/sem/ty/fn_ty.rs
xFrednet: The rename was mostly due to a clippy lint triggering. For ADT types, the `TyDefId` returned by `AdtTy::def_id()` is the id of the actual type, that it points to. Closures are similar, as they also point to a compiler generated type, which also has corresponding trait implementations. I'm not sure if it's better to name this `def_id()` which would indicate that this is similar to an ADT or something different, which better captures the kind of data this is and what it's used for :thinking:

Veetaha: I can't think of any ambiguity as in what `ty_id` or `def_id` would return for the closure, unless you think there may be a method on the closure that may return a different type def ID

xFrednet: Yeah, you're right. I now named it `def_id` to be inline with `AdtTy::def_id`.

Veetaha (approved):

Issue #324, opened 2023-11-26
[Bug]: `cargo uitest` fails with `[E0463]: can't find crate for `std``


tigerros:
### Summary

I'm not positive that this is a bug, but running Marker on a different crate instead of testing does work, so I think something is wrong with the testing.

### Reproducer

Windows 10

1. `git clone https://github.com/rust-marker/lint-crate-template.git`
2. `cd lint-crate-template`
3. `cargo uitest`

### Version

```text
cargo-marker 0.4.2
```

### Logs

```text
Updating crates.io index
   Compiling proc-macro2 v1.0.70
   Compiling unicode-ident v1.0.12
   Compiling serde v1.0.193
   Compiling cfg-if v1.0.0
   Compiling once_cell v1.18.0
   Compiling lazy_static v1.4.0
   Compiling windows_x86_64_msvc v0.48.5
   Compiling windows_x86_64_msvc v0.42.2
   Compiling winapi v0.3.9
   Compiling unicode-width v0.1.11
   Compiling sharded-slab v0.1.7
   Compiling tracing-core v0.1.32
   Compiling thread_local v1.1.7
   Compiling pin-project-lite v0.2.13
   Compiling serde_json v1.0.108
   Compiling windows-targets v0.48.5
   Compiling semver v1.0.20
   Compiling cc v1.0.83
   Compiling libc v0.2.150
   Compiling windows-sys v0.48.0
   Compiling quote v1.0.33
   Compiling windows-targets v0.42.2
   Compiling syn v2.0.39
   Compiling tracing v0.1.40
   Compiling tracing-subscriber v0.3.18
   Compiling thiserror v1.0.50
   Compiling portable-atomic v1.5.1
   Compiling regex-syntax v0.8.2
   Compiling ryu v1.0.15
   Compiling eyre v0.6.9
   Compiling crossbeam-utils v0.8.16
   Compiling anyhow v1.0.75
   Compiling itoa v1.0.9
   Compiling camino v1.1.6
   Compiling tracing-error v0.2.0
   Compiling regex-automata v0.4.3
   Compiling backtrace v0.3.69
   Compiling windows-sys v0.45.0
   Compiling rustc-demangle v0.1.23
   Compiling encode_unicode v0.3.6
   Compiling indenter v0.3.3
   Compiling owo-colors v3.5.0
   Compiling console v0.15.7
   Compiling color-spantrace v0.2.1
   Compiling serde_derive v1.0.193
[...]

PR #325, opened 2023-12-01
Rustc: Simplify `--sysroot` code after `rust#103660`

### LINKED ISSUE #48: Use rustc's way to get the system root
The driver like Clippy used to have a complicated way to find the system root for the compiler files. https://github.com/rust-lang/rust/pull/103660 added a new uniform way that should also be used by this rustc driver.

See: https://github.com/rust-linting/rust-linting/blob/ce1eb57dbdcf72f1b7ec820725ccf64bc2d90336/linter_driver_rustc/src/main.rs#L203

xFrednet:
If I'm honest, I still don't fully understand when and where the `--sysroot` argument is required. Clippy has very few comments on the topic. Miri has a few more, but also a custom setup in their `cargo-miri` interface, which is probably needed for the way miri operates.

Before https://github.com/rust-lang/rust/pull/103660, rustc used to use the path of the executable to get the system root, that's why Clippy and Miri had custom ways to search for the system root. When I wrote Marker's driver, I took Clippy's code for this part and refactored it for readability.

The referenced PR changed rustc's default behavior to use the path of the rustc library, which should work for Clippy, Miri and Marker. Clippy and Miri still have some minimal code, to check for an environment value. This PR mimics Clippy's behavior, with the deviation that we don't use `SYSROOT`, but `MARKER_SYSROOT` as the name of the environment value. I think this is a bit cleaner and inline with Miri, which uses `MIRI_SYSROOT`

---

Closes #48 

Maybe solves #324 as well? I sadly don't have a setup to replicate the reported bug. I'm considering, if we maybe want to do a test release and ask `@tigerros` (the author of #324) if they can test the release binaries for us. This PR has the v0.4.2 tag as the baseline, so we can just create a PR to the `hotfix/0.4` branch for that.

@Veetaha would you mind taking a look at this PR?

### DIFF
--- a/marker_rustc_driver/src/lib.rs
+++ b/marker_rustc_driver/src/lib.rs
@@ -37,9 +37,8 @@ pub mod lint_pass;
 
 use std::env;
 use std::ops::Deref;
-use std::process::Command;
 
-use camino::{Utf8Path, Utf8PathBuf};
+use camino::Utf8Path;
 use marker_adapter::{LintCrateInfo, LINT_CRATES_ENV};
 use marker_error::Context;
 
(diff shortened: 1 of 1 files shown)


Veetaha (approved):
I wouldn't worry about copying clippy's behavior. The less custom setup there is without breaking anything (at least in tests) the better.

xFrednet:
Yep agreed. Do you think we should do a patch version for this? I'm a little scared, that it might break things. But I'm guessing to 99% that it will be fine, and it might even fix the bug

Veetaha:
Let's try to do a patch `0.4.3-rc`. I remember there was some caveat with this with the toolchain version state. But since the toolchain version is currently the same as in the patch there shouldn't be any problem? Btw, even if the released version uses an older toolchain and we use the `hotfix` version with that toolchain in it it shouldn't cause any problems I think, but I may be misremembering

PR #326, opened 2023-12-02
Rustc: Simplify `--sysroot` code after `rust#103660` and prepare `0.4.3-rc`

xFrednet:
The first commit is the same as rust-marker/marker#325


CHANGELOG.md
```diff
@@ -24,10 +27,20 @@ The following components are considered to be internal and they are excluded fro
 
 ## [Unreleased]
 
+## [0.4.3] - 2023-12-02
```
Veetaha, line 30: Yeah, you can push it to `0.4.0-dev`. 
I also think this should be 
```suggestion
## [0.4.3-rc] - 2023-12-02
```

PR #340, opened 2023-12-18
API: Add `sem::TyKind::implements_trait` (Draft)

xFrednet:
This PR allows users to check if a given semantic type implements a specific trait. The implementation is inspired by Clippy's implementation of [`clippy_utils::ty::implements_trait`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/ty/fn.implements_trait.html). It already works for simple traits, without generics.

---

I'm stuck on how to allow users to specify generic arguments for the trait. My idea for Marker's function was that users can construct a `TestTraitRef`, which specifies a trait with its potential generics. I think this would make it a bit cleaner than Clippy's approach of taking a `trait_id: DefId` and `args: &[GenericArg<'tcx>]`. It also makes the interface more extendable. The problem is, that I can't figure out how to let the user specify generics in this way...

First some context. Generic arguments for the type check can be constructed from rustc's semantic types. Clippy constructs new semantic types for the type check, if needed. This is something which we'll also have to do in Marker, to support generics in types. Clippy doesn't distinguish between *real* types from the compiler and *fake* types constructed for this check. For Marker, I don't think this is a real option. If we allow users to construct `sem::TyKind`s I see numerous problems coming our way. Instead, I considered having a simplified type representation, that users can use to construct types for testing, something like this:

```
#[derive(Debug)]
enum TestType<'a> {
    Ref(&'a Test<'a>),
    Tuple(&'a [&'a Test<'a>]),
    Type(u32)
}
```

Then we could construct semantic types in the backend for the type check. And this is where the issue begins. I can't figure out a way to deal with the references in a good way, since I would like to store the thing inside the `TestTraitRef`. For that, we would have to allocate memory somewhere, as `'a` wouldn't live long enough otherwise.

Does anyone have a suggestion on how to best do this?

---

I can thing of three ways, which all feel a bit complicated:
1. Simply use an allocator in `marker_api` to extend the lifetime. This would add a dependency, but allow us to use these kinds of references.
2. Have a `Vec<TestType>` in the `TestTraitRef` that stores the nodes, extending their lifetimes to the life of `TestTraitRef`. This feels hacky and also doesn't solve all problems, as slices etc would need to be handled separately.
3. Use `Box<>` instead of references. Here I fear that we lock us in into a design, which can't be changed later. We could hide the usage of Boxes, by denying the user access to the `TestType` enum with its variants directly, and instead require them to use a builder of some kind. Using a builder is probably smart in either case. It still feels weird to me.

---

@Veetaha, do you maybe have an idea, how could this be solved? If you have the time, I'd appreciate it, if you could also take a look at the interface I currently have. Maybe you have some early feedback, which I can incorporate.

---

cc: rust-marker/marker#141


marker_api/src/common/id.rs
```diff
@@ -148,7 +148,7 @@ new_id! {
     /// This id is used by the driver to lint the semantic type representation, back to the
     /// driver type representation, if needed.
     #[cfg_attr(feature = "driver-api", visibility::make(pub))]
-    pub(crate) DriverTyId: u64
+    pub(crate) DriverTyId: u128
```
Veetaha, line 151: I wonder why this changed. I'd expect such breaking change only during the nightly toochain version update.

marker_api/src/sem/ty.rs
```diff
@@ -154,3 +203,183 @@ macro_rules! impl_ty_data {
...
+
+impl TestTraitRefBuilder {
+    fn new() -> Self {
+        Self {
+            trait_ref: None,
+            mode: TraitTestMode::default(),
+        }
+    }
+
+    pub fn trait_from_path(&mut self, path: impl Into<String>) -> &mut Self {
```
Veetaha, line 239: I think in general we should avoid allocations. 99% of the time the trait name will be a `&'static str`. So maybe this could be `impl Into<Cow<'_, str>>`

marker_api/src/sem/ty.rs
```diff
@@ -154,3 +203,183 @@ macro_rules! impl_ty_data {
...
+    }
+
+    pub fn mode(&mut self, mode: TraitTestMode) -> &mut Self {
+        self.mode = mode;
+        self
+    }
+
+    pub fn build(&mut self) -> TestTraitRef {
+        TestTraitRef {
+            trait_ref: self.trait_ref.take().expect("the trait was never set"),
```
Veetaha, line 256: I think [`buildstructor`](https://docs.rs/buildstructor/latest/buildstructor/) would be handy in this case since it validates such errors at compile time. Anyway, at some point we'll have to have a proc macro dependency in marker_api.

With buildstructor you could even achieve the API like this:

```rust
ctx.test_trait()
   .arg_foo(...)
   .arg_bar(...)
   .test()
```

In this case you won't even need any owned value as an `arg` like a `String`. Although with this approach it's harder to save the args into a variable and basically impossible to put them into some struct or enum because the builder type buildstructor generates is something very complicated and potentially breakable, so treat it as an anonymous type. However, I think this would be good enough

marker_api/src/sem/ty.rs
```diff
@@ -154,3 +203,183 @@ macro_rules! impl_ty_data {
...
+    }
+
+    pub fn mode(&self) -> TraitTestMode {
+        self.mode
+    }
+}
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum TyDefIdSource {
```
Veetaha, line 283: Does this need to be `pub`?

marker_api/src/sem/ty.rs
```diff
@@ -154,3 +203,183 @@ macro_rules! impl_ty_data {
...
+
+/// This enum defines how strict the [`TyKind::implements_trait`] test should be.
+/// The difference is probably best illustrated by examples. For that, let's
+/// take the following traits:
+///
+/// ```
+/// // Definitions
+/// trait SimpleTrait { /* ... */ }
+/// trait GenericTrait<T> { /* ... */ }
+/// trait TwoGenericTrait<T, U=u8> { /* ... */ }
```
Veetaha, line 309: There is also a case for const generics in traits, although I that's a rare case and we can implement support for that later

marker_api/src/sem/ty.rs
```diff
@@ -154,3 +203,183 @@ macro_rules! impl_ty_data {
...
+/// ```
+///
+/// Now we have a `SimpleType` which implements a few traits:
```
Veetaha, line 312: Maybe just squash this into one code snippet, and make this a line comment to duplicate code snippet a bit

marker_rustc_driver/src/context.rs
```diff
@@ -295,3 +331,33 @@ fn select_children_with_name(
...
+            param_env,
+            recursion_depth: 0,
+            predicate: mid::ty::Binder::dummy(trait_ref).to_predicate(self.rustc_cx),
+        };
+
+        // Let the record show: I'm not proud of this!!!
+
+        infcx
+            .evaluate_obligation(&obligation)
+            .is_ok_and(EvaluationResult::must_apply_modulo_regions)
```
Veetaha, line 361: I wonder if there is API in rustc that just returns the list of all traits that the type implements. For example, rustdoc can somehow display all the traits that the type implements both directly and via the blanket types.

If we just had a method that returned the full list then querying the trait implementations would require any special `TestTraitRef` DSL and other hussle. The users would be able to just iterate over the trait implementations manually. Ideally use iterators for that.

PR #342, opened 2023-12-26
Chore: Update Deps and increase MSRV to 1.70

xFrednet:
Not much to say, just some prep for the next release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ version = "0.5.0-dev"
 # endregion replace marker version dev
 
 # The MSRV is applied to the public library crates published to crates.io
-rust-version = "1.66"
+rust-version = "1.70"
 
 [workspace.dependencies]
 # region replace marker version dev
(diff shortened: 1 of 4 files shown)


xFrednet:
I believe the MSRV increase from 1.66 to 1.70 should be safe, since we distribute precompiled binaries now. We have several dependencies, that have 1.70 (release 2023-06-01) as their MSRV.

I still listed this change, in the changelog under the *breaking changes* section.

Cargo.toml
```diff
@@ -59,10 +59,10 @@ toml               = "0.8"
 tracing            = "0.1"
 tracing-error      = "0.2"
 tracing-subscriber = "0.3"
-typed-builder      = "0.16.2"
+typed-builder      = "0.18"
 ui_test            = "0.21.2"
 visibility         = "0.1.0"
-yansi              = "1.0.0-rc.1"
+yansi              = "1.0.0-gamma"
```
Veetaha, line 65: `1.0.0-rc.1` is a more recent one ([versions](https://crates.io/crates/yansi/versions))

Cargo.toml
```diff
@@ -44,9 +44,9 @@ marker_utils   = { path = "./marker_utils", version = "0.5.0-dev" }
 bumpalo            = "3.14"
 camino             = { version = "1.1", features = ["serde1"] }
 cargo_metadata     = "0.18"
-clap               = { version = "=4.3.24", features = ["string", "derive"] }
```
Veetaha, line 47: I wonder why it was pinned

Veetaha (approved):

Cargo.toml
xFrednet: Clap increased the minimal rust version in newer releases. The pin was me trying to keep rust 1.66 as the minimal version. Now we can just use the newest version again.

PR #345, opened 2023-12-29
Chore: Bump nightly -> 2023-12-28

xFrednet:
* So much breakage 14/27 fixed

* It compiles...


Veetaha:
I don't think `update-toolchain` script is needed here. Try adding this step before `cargo marker` invocation
```yml
      - uses: actions-rust-lang/setup-rust-toolchain@v1
```

